PR #1, opened 2018-03-01
Updates to dash-cluster-ansible

codablock:
1. Use "testnet" as network for sentinel
2. Support setting of custom sporkkey ids in devnet (requires https://github.com/dashpay/dash/pull/1960)
3. Fix full-disk problems due to huge docker log files
4. Drop cpuminer and rely on dashd + generate


roles/docker-options/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+docker_log_max_size: 256m
+docker_log_max_file: 4
```
nmarley: Needs EOL char

roles/dashd/templates/dash.conf.j2
```diff
@@ -65,3 +65,7 @@ addnode={{ dashd_seednode }}:{{ dashd_port }}
 masternode=1
 masternodeprivkey={{ masternode.privkey }}
 {% endif %}
+
+{% if dashd_sporkkeyid is defined %}
+sporkkeyid={{ dashd_sporkkeyid }}
```
nmarley: Note: Might need to update this arg depending on result of `sporkkeyid` discussion in Core PR (e.g. sporkaddr or whatever is settled on).

roles/docker-options/defaults/main.yml
codablock: fixed, but that is not shown here in GitHub

PR #9, opened 2018-08-18
[WIP] Evolution services and some fixes

shumkov:


nmarley:
@shumkov Could you please hold off on merging this one yet? I still want to make sure we don't include any identifying info in this and possibly squash locally (or use Squash Merge) if we have any in commits in here.

shumkov:
Yeah, sure. I have the same thought in my mind. But not sure about squashing. This branch kinda multi-branch (like master or develop) and would be cool to save commit history. What do you think?

nmarley:
Then I'd like to (possibly) manually rebase to ensure that any identifying config info isn't included in here. See also: https://12factor.net/config . This is a good methodology that I like to use as a guideline.

shumkov:
@nmarley You mean remove all commits with sensetive data? Probably we had some with our S3 backend configuration. I think that's all.
Yeah I always try to follow 12 factor principals. Love them.

PR #10, opened 2018-08-20
Insight configuration

shumkov:


cluster.yml
```diff
@@ -88,10 +88,13 @@
     - role: dashd-generate-miner
       when: dash_network != "mainnet"
 
-- hosts: faucet-web
```
nmarley: Why changing this back? This was changed to fix the conflicting group + host name. Or does this need a local pull + rebase first?

roles/insight/templates/dashcore-node.json
```diff
@@ -0,0 +1,25 @@
+{
+  "network": "{{ 'mainnet' if dash_network == 'mainnet' else 'testnet' }}",
```
nmarley: Shouldn't we just make this a variable? (Also, we can use a `.j2` extension if this is a standard Jinja2 template.)

nmarley: Reason is b/c neither of these options are valid if devnet, right? Or would that need more work in Insight?

terraform/aws/inventory/ansible_inventory.tpl
```diff
@@ -1,6 +1,6 @@
 ${all_hosts}
 
-[faucet-web]
```
nmarley: Same question as before, why reverting this?

terraform/aws/main.tf
```diff
@@ -61,6 +61,13 @@ resource "aws_security_group" "elb" {
     cidr_blocks = ["0.0.0.0/0"]
   }
 
+  ingress {
```
nmarley: Not sure why this is needed (the ELB should be forwarding from ELB port 80 to Insight host port 3001, so Insight nodes should not need to be accessible from the internet, just within the VPC subnet).

cluster.yml
shumkov: There are multifaucet and insight on these servers now. So I just renamed instances to `web-N` (as others) and brought back old group name.

roles/insight/templates/dashcore-node.json
shumkov: Insight doesn't know about devnet for now. And I think we don't need it yet. Devnet is kinda compatible with testnet, that's why we use testnet for all networks except mainnet. 
We will throw out from here Insight pretty soon when DAPI and wallet-library remove that dependency. 
Yes, we can make this variable, but this is temporary solution, so probably we should keep it for a while. What do you think?

terraform/aws/inventory/ansible_inventory.tpl
shumkov: https://github.com/dashpay/dash-networks-deploy/pull/10#discussion_r211407390

terraform/aws/main.tf
shumkov: 80 port is already busy by mutlifaucet :( I've added a task to TODO to use subdomains or paths instead of different ports.

roles/insight/templates/dashcore-node.json
nmarley: Ok, that's fine and we will throw this out anyway soon. 👍

terraform/aws/inventory/ansible_inventory.tpl
nmarley: 👍

cluster.yml
nmarley: Ok, makes sense if all web apps (e.g. faucet PHP + Insight) on same server.

terraform/aws/main.tf
nmarley: Ok. This usually isn't considered best practice and I think better to let the load balancer handle it. For now, I don't want to hold up progress. But moving forward can use Application ELB and use paths to specify the backend target (e.g. `/insight` could go to Insight and `/faucet` could go to PHP faucet).

PR #11, opened 2018-08-21
Use env variables for AWS ECR login

shumkov:


.env.example
```diff
@@ -1,4 +1,8 @@
+TERRAFORM_S3_BUCKET=""
+TERRAFORM_S3_KEY=""
+TERRAFORM_DYNAMODB_LOCK_TABLE=""
+
+AWS_ACCESS_KEY_ID=""
```
nmarley: Why do this?

awscli uses this if it's set (not sure what happens if it's set but empty), and falls back to AWS_PROFILE if it's set. I don't see what the point of setting these is (other than a suggestion). That's fine if we can be sure that it doesn't conflict with `AWS_PROFILE`. (We should check set but empty string.)

shumkov: AWS envs in `.env.example` just as the suggestion to set them. In your `.env` you can you `AWS_PROFILE` if you set it up.

PR #16, opened 2018-08-21
Add deploy/destroy bash script and cleanup project

shumkov:


networks/mainnet.yml
```diff
@@ -0,0 +1,71 @@
+---
+
+faucet_address: yhvXpqQjfN9S4j5mBKbxeGxiETJrrLETg5
```
nmarley: these are not mainnet addresses...

lib/utils.sh
```diff
@@ -0,0 +1,22 @@
+function print_error() {
+   echo "$1"
+   echo ""
+   echo "$CMD_USAGE"
+   exit 1
+}
+
+function override_aws_credentials() {
+    AWS_VARS="AWS_PROFILE AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS AWS_SECRET_ACCESS AWS_REGION"
```
nmarley: What is `AWS_SECRET_ACCESS` ? Env vars used by awscli are found here: https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html

lib/utils.sh
```diff
@@ -0,0 +1,22 @@
...
+
+    for VAR_NAME in ${AWS_VARS}; do
+        FIRST_VAR_NAME="$1_$VAR_NAME"
+        SECOND_VAR_NAME="$2_$VAR_NAME";
+
+        if [[ -n ${!FIRST_VAR_NAME} ]]; then
```
nmarley: You could just check for `$1` and `$2` at the top of the fcn & error out if both are not set.

.env.example
```diff
@@ -1,10 +1,25 @@
-TERRAFORM_S3_BUCKET="s3-bucket-name-here"
```
nmarley: Why remove these examples?

README.md
```diff
@@ -1,20 +1,23 @@
-# Dash Cluster Ansible
+# Dash Networks Deployment Tool
```
nmarley: probably "Network" makes more sense

README.md
```diff
@@ -1,20 +1,23 @@
...
 
+## Introduction
 
-## What is this?
+This tool allows deploy and manage Dash Networks networks.
```
nmarley: s/Networks networks/networks/

lib/utils.sh
nmarley: What is `AWS_SECRET_ACCESS` ? Environment variables that awscli uses are found here: https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html

lib/utils.sh
nmarley: Could probably check if `$1` and `$2` is set that the top of the function and error there instead.

networks/mainnet.yml
nmarley: These are not base58 encoded for mainnet.

.env.example
shumkov: It might mislead users like it's already set and they just skip configuration. And then quite complicated to find a mistake. If it's empty it's more obvious what is the problem. I faced that before.

shumkov: I think it's better to add comments with examples if needed

README.md
shumkov: I thought if we are deploying multiple networks we should keep it plural. No?
If I'm wrong we should rename repo too.

.env.example
nmarley: Ok, I take your point. Maybe we can specify instructions (link to README if not set) like you did somewhere else in this project.

README.md
nmarley: It's per-network. Makes more sense in English to call it a "network deployment tool" -- it is understood that the tool can be used over & over. But you can call it what you want.

shumkov: Ok. I see. Then I will fix it everywhere.

.env.example
shumkov: Done

lib/terraform.sh
```diff
@@ -12,10 +12,11 @@ function run_terraform_command() {
     # Initialize terraform w/remote AWS backend configuration
 
     terraform init \
-        -backend-config="bucket=$TERRAFORM_S3_BUCKET" \
-        -backend-config="key=$TERRAFORM_S3_KEY" \
+        -backend="$TERRAFORM_BACKEND"
```
nmarley: For one, I think this PR is too large and would prefer if backend setup moved to another PR.

Also, don't think this is the best way... I think we should default to "local" instead of skipping backend config altogether.

See: https://www.terraform.io/docs/commands/init.html#backend-initialization

> To skip backend configuration, use -backend=false. Note that some other init steps require an initialized backend, so it is recommended to use this flag only when the working directory was already previously initialized for a particular backend.

PR #17, opened 2018-08-24
Add Dash Drive and MongoDB to mn-services Ansible role

dreyacosta:


ansible/roles/mn-services/templates/docker-compose.yml
```diff
@@ -24,9 +24,47 @@ services:
...
+  drive_sync:
+    image: 103738324493.dkr.ecr.us-west-2.amazonaws.com/dashevo/dashdrive:latest
+    restart: always
+    command: npm run sync
+    environment:
+      - DASHCORE_JSON_RPC_USER={{ dashd_rpc_user }}
+      - DASHCORE_JSON_RPC_PASS={{ dashd_rpc_password }}
+      - DASHCORE_JSON_RPC_HOST={{ private_ip }}
+      - DASHCORE_JSON_RPC_PORT={{ dashd_rpc_port }}
+      - STORAGE_MONGODB_DB=drive_main
```
shumkov: Why we can't use the default value?

dreyacosta: You're right, I forgot to remove that after a few checks 👍

PR #18, opened 2018-08-24
Use verbose tags for AWS resources

shumkov:


terraform/aws/main.tf
```diff
@@ -9,26 +9,46 @@ data "aws_ami" "ubuntu" {
   most_recent = true
 
   filter {
-    name   = "name"
-    values = ["ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*"]
+    name = "name"
+
+    values = [
+      "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*",
```
nmarley: Just a note: we could even try 18.04, but 16.04 is probably ok too since it's LTS.

Prefix in case you wanted to test it : `ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-`

shumkov:
Yeah man, Web Storm does it automatically before commit.

PR #21, opened 2018-08-29
Document Ansible arguments for Evolution's services

dreyacosta:


.env.example
```diff
@@ -30,4 +30,20 @@ AWS_REGION=""
 # ANSIBLE_AWS_REGION=""
 
 # Additional ansible-playbook arguments
+#
```
shumkov: I think we should write it in README in addition section. Mmm?

README.md
```diff
@@ -48,6 +48,17 @@ cp .env.example .env
 $EDITOR .env
 ```
 
+### Evolution configuration
+Several Ansible's arguments should be specified in `.env` file to deploy Evolution's services:
+```bash
+ANSIBLE_ARGUMENTS="-e aws_ecr_login=true \
```
shumkov: `aws_ecr_login` is optional only if you are using AWS ECR for your images

PR #30, opened 2018-09-16
Smoke tests

abvgedeika:
how to run:
npm install
./bin/test devnet-example -t=smoke -m=" --timeout 10000"


test/smoke/dapi.js
```diff
@@ -1,10 +1,38 @@
-describe('DAPI', () => {
```
shumkov: Let's keep root 'DAPI' like:
```
- DAPI
   - masternode-1
      - should respond data from chain
```

test/smoke/dapi.js
```diff
@@ -1,10 +1,38 @@
...
+  const options = {
+    debug: false,
+    verbose: false,
+    errors: false,
+    warnings: false,
+    seeds: [{ ip: networkConfig.inventory._meta.hostvars[nodeName].public_ip }],
+  };
+  describe(`DAPI ${nodeName}`, () => {
+    // getBlockHash height:1
+    it('should respond data from chain', async () => {
```
shumkov: How do you think should we compare obtained data with data from DC just to make sure that DAPI connected well to correct node?

test/smoke/dashCore.js
```diff
@@ -1,18 +1,79 @@
...
-  });
-  describe('Miners', () => {
-    // waitfornewblock
-    it('should mine blocks');
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
+
+networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
```
shumkov: Something wrong with stucture here. "All nodes" should test all DC nodes, not just masternodes: wallet-nodes + full-nodes + miners + masternodes. Miners should test only miners.

test/smoke/faucet.js
```diff
@@ -1,4 +1,27 @@
...
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
+
 describe('Faucet', () => {
-  // getFunds
-  it('should respond positive balance');
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
```
shumkov: Facets located on `web` nodes.

test/smoke/dapi.js
```diff
@@ -1,10 +1,38 @@
...
+      const dapi = new DAPIClient(options, 3000);
```
shumkov: Why do you create client each time? You can create on for each node in `before` section.

test/smoke/dashCore.js
```diff
@@ -1,18 +1,79 @@
...
+      protocol: 'http',
+      user: 'dashrpc',
+      pass: 'password',
+      host: networkConfig.inventory._meta.hostvars[nodeName].public_ip,
+      port: 20002,
+    };
+    const rpc = new RpcClient(config);
+
+    describe('All nodes', () => {
+      it('should have correct network type/subversion from `getNetworkInfo`', async () => {
```
shumkov: I think we should keep specs less technical. "should have correct network type" will be enought.

test/smoke/dashCore.js
```diff
@@ -1,18 +1,79 @@
...
+        expect(blockHash.result.networkactive).to.be.equal(true);
+        expect(blockHash.result.subversion).to.be
+          .equal(`/Dash Core:${version}(${networkConfig.network.type}=${networkConfig.network.name})/`);
+      });
+
+      // The number of blocks should be almost the same (-3/+3) and block hash of particular
+      // block height should be the same. Using `blocks`
+      // and `bestblockhash` from `GetBlockChainInfo`.
+      it('should have correct blockhash and blocks count', async () => {
+        const bestBlockHash = await rpc.getBestBlockHash();
```
shumkov: Why do we compare RPC response with cli? I don't understand the meaning. The idea was to compare values from all nodes and the average should be the same.

test/smoke/dashCore.js
```diff
@@ -1,18 +1,79 @@
...
+        for (const nName of networkConfig.inventory.masternodes.hosts) {
+          masterIpsInv.push(networkConfig.inventory._meta.hostvars[nName].public_ip);
+        }
+        expect(masterIps.sort()).to.deep.equal(masterIpsInv.sort());
+      });
+    });
+    describe('Miners', () => {
+      // waitfornewblock
+      it('should mine blocks', async function () {
+        this.timeout(160000);
```
shumkov: You can use `waitfornewblock` blocking RPC call

test/smoke/faucet.js
```diff
@@ -1,4 +1,27 @@
...
+    it('should respond positive balance via insight and cli',
+      async () => {
+        const response = await utils.requestWrapper(`http://${networkConfig.inventory._meta.hostvars[nodeName].public_ip}:3001/insight-api-dash/addr/${networkConfig.variables.faucet_address}`);
```
shumkov: I was thinking we just call faucet service on 80 port. `getFunds` RPC call.

test/testUtils.js
```diff
@@ -0,0 +1,65 @@
...
+      error(err) {
+        error = err;
+        reject(new Error(`exit code: ${code}`));
+      },
+    })
+      .start();
+  }));
+};
+
+exports.requestWrapper = function (url) {
```
shumkov: Let's use https://github.com/bitinn/node-fetch instead of this wrapper. It supports promises and many more.

test/smoke/dashCore.js
shumkov: Ops, probably we need to add it. I'll do that tomorrow.

test/smoke/dapi.js
```diff
@@ -3,36 +3,43 @@ const getNetworkConfig = require('../../lib/test/getNetworkConfig');
...
-      const blockHash = await dapi.getBlockHash(1);
-      expect(blockHash).to.be.an('string');
-    });
+const fetch = require('node-fetch');
 
-    // getBestBlockHeight
-    it('should respond data from insight', async () => {
+describe('DAPI', () => {
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
+    describe(`DAPI ${nodeName}`, () => {
```
shumkov: Maybe just node name will be enough?

test/smoke/dapi.js
```diff
@@ -1,10 +1,45 @@
...
+        errors: false,
+        warnings: false,
+        seeds: [{ ip: networkConfig.inventory._meta.hostvars[nodeName].public_ip }],
+      };
+      const dapi = new DAPIClient(options, 3000);
+      // getBlockHash height:1
+      it('should respond data from chain', async () => {
+        const blockHash = await dapi.getBlockHash(1);
+        // const response = await utils.requestWrapper(`http://${networkConfig.inventory._meta.hostvars[nodeName].public_ip}:3001/insight-api-dash/block-index/1`);
+        const resp = await fetch(`http://${networkConfig.inventory._meta.hostvars[nodeName].public_ip}:3001/insight-api-dash/block-index/1`)
```
shumkov: Why not directly from DashCore? As I know we are going to get rid of Insight soon.

test/smoke/dapi.js
```diff
@@ -1,10 +1,45 @@
...
+        const blockHeight = await dapi.getBestBlockHeight(1);
+        expect(blockHeight).to.be.an('number');
+        expect(blockHeight % 1).to.be.equal(0);
+      });
 
-  // fetchDapObjects any dapiId and type
-  it('should respond data from drive');
+      // fetchDapObjects any dapiId and type
+      it('should respond data from drive', async () => {
+        const balance = await dapi.getBalance(`${networkConfig.variables.faucet_address}`);
```
shumkov: It supposes to make a request to drive?

test/smoke/dashCore.js
```diff
@@ -1,18 +1,116 @@
...
+        pass: 'password',
+        host: networkConfig.inventory._meta.hostvars[nodeName].public_ip,
+        port: 20002,
+      };
+      const rpc = new RpcClient(config);
+
+      it('should have correct network type', async () => {
+        const blockHash = await rpc.getNetworkInfo();
+        const masternodelist = await rpc.masternodelist();
+        const version = masternodelist.result[Object.keys(masternodelist.result)[0]].daemonversion;
```
shumkov: I think checking version is unnecessary. Just check that it includes substring: `(${networkConfig.network.type}=${networkConfig.network.name})`

test/smoke/dashCore.js
```diff
@@ -1,18 +1,116 @@
...
+const fs = require('fs');
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
+
+describe('All nodes', () => {
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
+    describe(`DashCore ${nodeName}`, () => {
```
shumkov: Just node name?

test/smoke/dashCore.js
```diff
@@ -1,18 +1,116 @@
...
+const SSH = require('simple-ssh');
+const fs = require('fs');
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
+
+describe('All nodes', () => {
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
```
shumkov: We should iterate over all dashd nodes: https://github.com/dashpay/dash-network-deploy/pull/30#discussion_r217921683

test/smoke/dashCore.js
```diff
@@ -1,18 +1,116 @@
...
+        protocol: 'http',
+        user: 'dashrpc',
+        pass: 'password',
+        host: networkConfig.inventory._meta.hostvars[hosts[i]].public_ip,
+        port: 20002,
+      };
+      const rpc = new RpcClient(config);
+
+      const bestBlockHash = await rpc.getBestBlockHash();
+      const blockCount = await rpc.getBlockCount();
```
shumkov: I think here `getBlockchainInfo` will be enough. We don't need to compare it with `getBestBlockHash` and  `getBlockCount` with `getBlockchainInfo`. They are always will be equal.

But we need to check hashes on particular height. So In addition to `blockCounts` we need to aggregate block hashes by heights and make sure that they are equal. For example:
```
if (!blockhashes[blockchainInfo.result.blocks]) {
   blockhashes[blockchainInfo.result.blocks] = blockchainInfo.result.bestblockhash;
}
expect(blockhashes[blockchainInfo.result.blocks]).to.be.equal(blockchainInfo.result.bestblockhash);
```

test/smoke/faucet.js
```diff
@@ -1,4 +1,23 @@
...
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
 describe('Faucet', () => {
-  // getFunds
-  it('should respond positive balance');
+  networkConfig.inventory.web.hosts.forEach((nodeName) => {
+    it('should respond positive balance via insight and cli',
```
shumkov: Let's add `describe` with node name here too.

test/smoke/faucet.js
```diff
@@ -1,4 +1,23 @@
...
+      async () => {
+        const ssh = new SSH({
```
shumkov: Oh... faucet has RPC API which we can call with `fetch`: http://www.unibia.com/unibianet/developer/multifaucet-easy-setup-crypto-currency-faucet-supports-multiple-currencies

test/smoke/insight.js
```diff
@@ -1,4 +1,24 @@
...
+
+const networkConfig = getNetworkConfig();
+
+const fetch = require('node-fetch');
+
+networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
+  describe(`Insight ${nodeName}`, () => {
+    it('should return block',
+      async () => {
+        const resp = await fetch(`http://${networkConfig.inventory._meta.hostvars[nodeName].public_ip}:3001/insight-api-dash/blocks?limit=1`)
```
shumkov:
```
const response = await fetch(...);
const { body: { blocks }}  = await response.json();
expect(blocks).to.be.equal(1);
```

test/smoke/ipfs.js
```diff
@@ -1,4 +1,16 @@
...
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
 describe('IPFS', () => {
   // swarm peers
-  it('should be interconnected');
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
+    it('should be interconnected', async () => {
```
shumkov: Let's add `describe` with node name here too

test/testUtils.js
shumkov: We don't use `requestWrapper` anymore. Let's remove it.

test/smoke/dapi.js
abvgedeika: do you mean 'DAPI nodeName'? it doesn't work

test/smoke/dapi.js
shumkov: You can call `getDapContract` or `getUserDapSpace`.

test/smoke/dashCore.js
shumkov: Doesn't work like this `describe(nodeName, () => {`

package.json
```diff
@@ -21,6 +21,9 @@
   },
   "homepage": "https://github.com/dashpay/dash-network-deploy#readme",
   "devDependencies": {
+    "@dashevo/dapi-client": "git+ssh://git@github.com/dashevo/dapi-client.git",
+    "@dashevo/dashd-rpc": "git+ssh://git@github.com/dashevo/dashd-rpc.git#develop",
+    "jayson": "^2.0.6",
```
shumkov: I think we don't need `jayson`  coz we don't use it at all.

test/smoke/dapi.js
```diff
@@ -1,10 +1,51 @@
...
+        pass: 'password',
+        host: networkConfig.inventory._meta.hostvars[nodeName].public_ip,
+        port: 20002,
+      };
+      const rpc = new RpcClient(config);
+      // getBlockHash height:1
+      it('should respond data from chain', async () => {
+        const blockHash = await dapi.getBlockHash(1);
+        const blockHashRpc = await rpc.getBlockHash(1);
+        expect(blockHash).to.be.an('string');
```
shumkov: What's the point of this check?

test/smoke/dashCore.js
```diff
@@ -1,18 +1,114 @@
...
+        user: 'dashrpc',
+        pass: 'password',
+        host: networkConfig.inventory._meta.hostvars[nodeName].public_ip,
+        port: 20002,
+      };
+      const rpc = new RpcClient(config);
+
+      it('should have correct network type', async () => {
+        const networkInfo = await rpc.getNetworkInfo();
+        expect(networkInfo.error).to.be.equal(null);
```
shumkov: I thought RPC throws an error if can't request to DC instance?

test/testUtils.js
```diff
@@ -0,0 +1,52 @@
+const SSH = require('simple-ssh');
+const fs = require('fs');
+const Promise = require('bluebird');
+
+exports.echo = function (ssh, command, callback) {
```
shumkov: Why do you not use promises here?

test/smoke/ipfs.js
```diff
@@ -1,4 +1,16 @@
...
+
+const getNetworkConfig = require('../../lib/test/getNetworkConfig');
+
+const networkConfig = getNetworkConfig();
+
 describe('IPFS', () => {
   // swarm peers
-  it('should be interconnected');
+  networkConfig.inventory.masternodes.hosts.forEach((nodeName) => {
+    it(`should be interconnected ${nodeName}`, async () => {
```
shumkov: `describe(nodeName)`?

test/smoke/dapi.js
abvgedeika: there are fetchDapContract &  fetchDapObjects instead of getDapContract/getUserDapSpace but they need valid dapId in params. so, to fetch by dapId firstly we need create this object. I guess it's not a sanity test and should be part of integration scenarios.

test/smoke/dapi.js
abvgedeika: only to be sure that blockHash is not undefined. will remove it

test/smoke/dashCore.js
abvgedeika: getNetworkInfo returns
```
{ result: 
   { version: 120400,
     subversion: '/Dash Core:0.12.4(devnet=devnet-example)/',
     protocolversion: 70211,
     localservices: '0000000000000005',
     localrelay: true,
     timeoffset: 0,
     networkactive: true,
     connections: 9,
     networks: [ [Object], [Object], [Object] ],
     relayfee: 0.00001,
     incrementalfee: 0.00001,
     localaddresses: [ [Object] ],
     warnings: 'This is a pre-release test build - use at your own risk - do not use for mining or merchant applications' },
  error: null,
  id: 13122 }
```
I only check that error is null here

test/smoke/dapi.js
shumkov: No worries, those methods respond you that they don't have these objects/contract but that's enough for us to make sure that services are working and connected.

test/smoke/dashCore.js
shumkov: I see, ok. Then you need to check error for every JSON RPC call or use jayson client (what I recommend).

test/smoke/dapi.js
abvgedeika: I see that fetchDapObjects  has been added 5 days ago https://github.com/dashevo/dapi-client/blob/master/src/index.js . when I call it I get 
Error: DAPI RPC error: fetchDapContract: Method not found
      at Object.request (node_modules/@dashevo/dapi-client/src/RPCClient.js:30:11)

do we need to update dapi instances?

shumkov: Yep, @antohou please push new changes.

test/smoke/dapi.js
andrei-baranouski: I guess that dashCore.'All Nodes' verify consistency of this data

test/smoke/dashCore.js
andrei-baranouski: done. though rpc reuests to miner hosts stuck now. any ideas why it could be?

test/smoke/faucet.js
andrei-baranouski: fixed

test/smoke/dapi.js
andrei-baranouski: fixed

test/smoke/dashCore.js
andrei-baranouski: fixed

test/smoke/dashCore.js
andrei-baranouski: fixed. now compare only rpc responses between all nodes.

test/smoke/dashCore.js
andrei-baranouski: NA

test/smoke/faucet.js
andrei-baranouski: we were not able to get faucet via rpc.getFunds. only check balance by direct IP:80/ request

test/testUtils.js
andrei-baranouski: changed

test/smoke/dapi.js
andrei-baranouski: fixed

test/smoke/dapi.js
andrei-baranouski: now use rpc.getBlockHash(1)

test/smoke/dapi.js
andrei-baranouski: dapi image has been updated. now dashdrive lib should be updated

test/smoke/dashCore.js
andrei-baranouski: fixed

test/smoke/dapi.js
shumkov: Hm, right.

test/smoke/dashCore.js
shumkov: Because of problems with new version of DC. Just ignore it for a while please.

PR #36, opened 2018-09-21
Configure DAPI service

shumkov:


shumkov:
Not sure that we can pass devnet as network name. Gonna deploy it for testing first.

PR #48, opened 2018-10-22
Check if drive is synced while 1.5 minutes

abvgedeika:


test/smoke/drive.js
```diff
@@ -23,11 +25,19 @@ describe('Drive', () => {
...
-        this.slow(1000);
-
-        const { result: info } = await driveClient.request('getSyncInfo', {});
-
-        expect(info).to.have.property('lastSyncedBlockHeight');
+        this.timeout(100000);
+        this.slow(100000);
+
+        const time = Date.now() + 90000;
+        while (time > Date.now()) {
```
shumkov: Don't we use just for loop with a counter:
```
const timeout = 1000;
const attempts = 90;
for (const i = 0; i <= attempts; i++) {
     ...
     await wait(timeout);
}
```

abvgedeika: with for loop there potential problems:
1)requests inside loop could be proceed a long time. in this case 90 requests could take even hours
2) we are not able to track/calculate properly this.timeout and this.slow mocha methods

test/smoke/drive.js
```diff
@@ -23,11 +25,19 @@ describe('Drive', () => {
...
-
-        expect(info).to.have.property('lastSyncedBlockHeight');
+        const timeout = 1000;
+        const attempts = 90;
+        this.timeout((attempts + 10) * timeout);
+        this.slow((attempts + 10) * timeout);
+        for (let i = 0; i <= attempts; i++) {
+          const { result: info } = await driveClient.request('getSyncInfo', {});
+          if (info.status === 'synced') {
+            expect(info).to.have.property('lastSyncedBlockHeight');
```
shumkov: We don't need this check since we are checking `info.status`

test/smoke/drive.js
```diff
@@ -23,11 +25,18 @@ describe('Drive', () => {
...
-
-        const { result: info } = await driveClient.request('getSyncInfo', {});
-
-        expect(info).to.have.property('lastSyncedBlockHeight');
+        const timeout = 1000;
+        const attempts = 90;
+        this.timeout((attempts + 10) * timeout);
+        this.slow((attempts + 10) * timeout);
+        for (let i = 0; i <= attempts; i++) {
+          const { result: info } = await driveClient.request('getSyncInfo', {});
```
shumkov: We have to handle an error too.
```
const { result, error } = await ...
if (error) {
 except.fail(error.message);
}
...
```

test/smoke/drive.js
```diff
@@ -18,16 +20,27 @@ describe('Drive', () => {
       });
 
       it('should respond current sync status', async function it() {
-        if (!variables.evo_services) {
-          this.skip('Evolution services are not enabled');
-          return;
+        // if (!variables.evo_services) {
```
shumkov: Please bring it back

test/smoke/drive.js
```diff
@@ -18,16 +20,27 @@ describe('Drive', () => {
...
+        // }
+
+        const timeout = 1000;
+        const attempts = 90;
+        this.timeout((attempts + 10) * timeout);
+        this.slow((attempts + 10) * timeout);
+        for (let i = 0; i <= attempts; i++) {
+          const { result: info } = await driveClient.request('getSyncInfo', {});
+          if (info.status === 'synced') {
+            const { _, error } = await driveClient.request('getSyncInfo', {});
```
shumkov: Why we are doing second request? Don't we check errors for the first one?

PR #49, opened 2018-11-01
fix 'should respond data from drive' test

abvgedeika:


shumkov:
That’s already fixed.

PR #59, opened 2018-12-25
Pass absolute path to Docker image entrypoint

shumkov:


nmarley:
As an aside, this Dockerfile w/install && then uninstall of build tools looks like it might be a good candidate for a [multi-stage build Dockerfile](https://docs.docker.com/develop/develop-images/multistage-build/) instead.

PR #63, opened 2019-01-20
Fix provisioning destruction

shumkov:


ansible/roles/kill-network/tasks/main.yml
```diff
@@ -21,7 +21,7 @@
   shell: docker volume ls -qf dangling=true | xargs -r docker volume rm
 
 - name: remove all docker images
-  shell: docker images --no-trunc | awk '{ print $3 }' | xargs -r docker rmi
+  shell: docker image ls -q | awk '{ print $3 }' | xargs -r docker image rm
```
nmarley: What is the point of changing `images` to `image` and `rmi` to `image rm`?

The only thing which actually changes here is the `-q` argument instead of vs `--no-trunc`, which changes the output. Also, this won't work with the `awk` command anymore, because this will just show blanks as there is no 3rd field. You can remove the `awk` pipe entirely I think.

shumkov:
> What is the point of changing images to image and rmi to image rm?

As far I understand `docker images` and `docker rmi` will be deprecated in the future.

nmarley:
> As far I understand docker images and docker rmi will be deprecated in the future.

I can't find this anywhere. Can you point me to any relevant documentation?

shumkov:
@nmarley that's my guess. They introduced CLI command section due to the increasing number of commands, so I think it makes sense to use new format than old which is literally legacy.

PR #67, opened 2019-01-24
Fix IPFS 0.4.18 get peer id workaround

shumkov:


nmarley:
Something like this might work w/o the lock issue, but your exec command is still better. Just another option in case this continues:

`docker-compose exec -T ipfs cat /data/ipfs/config | jq -r .Identity.PeerID`

Goes thru the filesystem instead of ipfs

PR #69, opened 2019-01-31
Add `dash-cli` command to `dash-network` tool

shumkov:


nmarley:
I think we should rebase after merging #68 , right @shumkov ? Seems like I just reviewed some of these changes there.

shumkov:
@nmarley Yep, that's right. I used `logs-command` as a base branch in order to use `ansible_get_ip_by_host` and `openvpn_start` functions.

bin/dash-cli
```diff
@@ -0,0 +1,63 @@
...
+    testnet)
+        NETWORK_ARG="-testnet"
+    ;;
+    regtest)
+        NETWORK_ARG="-regtest"
+    ;;
+esac
+
+dash-cli -rpcconnect="$ANSIBLE_HOST" \
+         -rpcport=20002 \
```
nmarley: 🤔 Are we still using this hard-coded port number everywhere, for all networks?

shumkov: Actually, it is not hardcoded. You can overwrite it for your deployment:

https://github.com/dashpay/dash-network-deploy/blob/master/ansible/group_vars/all#L35
https://github.com/dashpay/dash-network-deploy/blob/master/terraform/aws/variables.tf#L8

shumkov:
@nmarley I've updated readme

bin/dash-cli
shumkov: @nmarley Yep, I was going to read configuration in the next iteration. Probably I need to add comments. 

Actually, I am thinking about rewriting these bash scripts in Go.

nmarley: > Actually, I am thinking about rewriting these bash scripts in Go.

nmarley: > Actually, I am thinking about rewriting these bash scripts in Go.

Hah, I actually thought about that this weekend! 😄  But I'm not certain that it will be worth the effort right now, since these tools work "good enough" for now and are lower priority than other things.

PR #77, opened 2019-03-27
Update Insight API

shumkov:


antouhou:
Hm, what is the difference between reviewers and assignees?

shumkov:
No idea.

ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -44,6 +44,7 @@ rpcallowip=0.0.0.0/0
 rpcallowip={{ dashd_private_cidr }}
 rpcallowip=127.0.0.0
 rpcallowip=172.0.0.0/8
```
v2nek: Here should be 172.16.0.0/12 as rest of network is routable to real internet

shumkov: Sometimes docker interface is `172.17...` or even different, no?

nmarley: Since this runs within Docker I'd recommend setting this (only) to `0.0.0.0/0` and letting the host handle the infra. Networking within Docker doesn't mean that this is available outside the Docker network, only to all hosts in the network this container is running.

ansible/roles/mn-evo-services/templates/dashcore-node.json.j2
```diff
@@ -0,0 +1,24 @@
+{
+  "network": "{{ 'mainnet' if dash_network == 'mainnet' else 'testnet' }}",
```
nmarley: Why not just `{{ dash_network }}` then?

ansible/roles/mn-evo-services/templates/docker-compose.yml.j2
```diff
@@ -79,10 +79,28 @@ services:
...
+  insight_api:
+    image: {{ insight_image }}
+    container_name: insight_api
+    restart: always
+    ports:
+      - 3002:3001
+    volumes:
+      - {{ mn_evo_services_path }}/dashcore-node.json:/root/.dashcore/dashcore-node.json
+
+  # in order to support obsolete URL path (/insight-api-dash)
```
nmarley: Do we know what is still using this old path, and can we change those instead?

ansible/roles/dashd/templates/dash.conf.j2
shumkov: It runs using host network https://github.com/dashpay/dash-network-deploy/blob/48c535806218310a4a3bb12f3b538175401ae9b2/ansible/roles/dashd/tasks/main.yml#L52

ansible/roles/mn-evo-services/templates/dashcore-node.json.j2
shumkov: It doesn't support "devnet"

ansible/roles/mn-evo-services/templates/docker-compose.yml.j2
shumkov: Yes, we could but how we will force our users to update this software?

nmarley: Easy -- by not serving this pass-thru route and forcing them to change their paths. This is a band-aid fix, and they should be updating, instead of us providing this band-aid, which:

1) is extra complexity for us, and
2) provides no incentive for anyone to ever change the path to the new Insight API path (which has been around for a while now)

Anyway, this isn't supposed to serve the entire internet via Insight, right? This is just for internal use for DAPI? Or under which scenarios is Insight being used in this repo?

ansible/roles/mn-evo-services/templates/dashcore-node.json.j2
nmarley: Ok, thanks. Good to know.

ansible/roles/mn-evo-services/templates/docker-compose.yml.j2
shumkov: Honestly, I do not like this strategy and "break compatibility" without any announcements. 

In this setup, We are trying to reproduce the mainnet, so Insight API is not just for DAPI, but for other 3rd party needs.

nmarley: It will never be fixed if you don't fix it now. Honestly I don't think there's anyone using this but us, so I don't understand your argument for not fixing this now before the "band-aid" gets widespread. Then people really will start using the wrong path.

I think we have an opportunity to get this right early on, and we're doing something the wrong way by adding extra complexity and adding extra paths which really shouldn't exist anyway. I would be _very_ surprised if anyone is using this path at all right now. And this has been gone in Insight-API for months.

Can you tell me, what users do you think are using this to deploy Insight, anywhere? And what consumers (apps, etc.) are using this `insight-api-dash` path? I don't think there are any, but maybe you know of some that I am not aware of.

v2nek: @nmarley i've tried to remove this endpoint, and there were problems in mobile clients.
I think that we should develop strategy to deprecate endpoint and this is out of scope of this pull request. Current setup is like this and we should maintain it. If we need to remove this endpoint (which i agree we need), so we should launch some deprecation process to get rid of this url, which will involve official announcement in the company, outside of the company, measuring number of current requests to this endpoint and comparing number of requests over time. And in few month after this process started we can safely remove this endpoint, because everybody who wanted - already updated their software to use actual endpoint.

nmarley: Ok, makes sense. Let's do that then.

PR #85, opened 2019-06-11
Introduce DAPI GRPC Core service

shumkov:


ansible/roles/mn-evo-services/templates/docker-compose.yml.j2
```diff
@@ -67,10 +67,12 @@ services:
     depends_on:
       - drive_api
     ports:
-      - 3004:3000
+      - 3004:3004
```
abvgedeika: outside/public dapi port should be 3000, right?
oh, I see that public port in nginx is 80 and then requests are scattered around the ports

PR #90, opened 2019-07-31
Use multiple subnets for deploy

nmarley:
- adds a blacklist for AZs known to be not working
- renames the "private" subnets to public (they were always public because of
  the default route table association w/a route to the internet)
- adds a list variable for subnet CIDR ranges
- use the different subnets instead of locating all instances into one single
  AZ


nmarley:
This uses HCL introduced in Terraform v0.12 and requires #89 to be merged first!

I can rebase if needed once that PR is first approved / merged.

PR #91, opened 2019-07-31
Update `kyl191.openvpn` ansible role

shumkov:


nmarley:
@shumkov Have you verified this is working now or do we still need to make some changes to the role?

shumkov:
Doesn’t work yet. I’ve sent you details.

shumkov:
@nmarley Now it's ok. We can merge it

PR #101, opened 2019-12-13
feat: parametrize faucet via env vars

v2nek:


ansible/roles/multifaucet/templates/multifaucet/faucet.conf.php
```diff
@@ -3,10 +3,10 @@
 $config = array(
         "mysql_table_prefix" => getenv("MULTIFAUCET_DB_PRFX"), // table prefix to use
 
-        "minimum_payout" => 1000, // minimum to be awarded
-        "maximum_payout" => 1000, // maximum to be awarded
-        "payout_threshold" => 1000, // payout threshold, if the faucet contains less than this, display the 'dry_faucet' message
-        "payout_interval" => "1m", // payout interval, the wait time for a user between payouts. Type any numerical value with either a "m" (minutes), "h" (hours), or "d" (days), attached. Examples: 50m for a 50 minute delay, 7h for a 7 hour delay, etc.
+        "minimum_payout" => getenv("MULTIFAUCET_MIN_PAYOUT"), //1000, // minimum to be awarded
```
shumkov: Let's keep default value if env var is empty

v2nek: It is "defaulted" from ansible code, and we are not planning to run it outside of ansible.
Anyway is there a simple way to do that in php?

shumkov: I would do that on both sides. Just use ternary operator and if env variable is empty.

Since PHP 5.3 you can do:
```php
"minimum_payout" => getenv("MULTIFAUCET_MIN_PAYOUT") ?: 1000, 
```

ansible/group_vars/all
```diff
@@ -72,6 +72,10 @@ faucet_rpc_port: '{{ dashd_rpc_port }}'
 faucet_rpc_user: '{{ dashd_rpc_user }}'
 faucet_rpc_password: '{{ dashd_rpc_password }}'
 faucet_rpc_args: '-rpcconnect={{ faucet_rpc_host }} -rpcport={{ faucet_rpc_port }} -rpcuser={{ faucet_rpc_user }} -rpcpassword={{ faucet_rpc_password }}'
+faucet_min: 10
```
shumkov: faucet_min_payout
faucet_min_payout
faucet_payout_threshold
faucet_payout_interval

PR #109, opened 2019-12-30
feat: introduce envoy proxy for gRPC Web

shumkov:


ansible/roles/mn-evo-services/templates/dapi-nginx.conf.j2
```diff
@@ -101,6 +107,8 @@ server {
         }
 
         proxy_pass http://{{ private_ip }}:3020$request_uri;
+        proxy_http_version 1.1;
+        proxy_hide_header access-control-allow-origin;
```
shumkov: Sorry, I don't get how it works. What is the point to add `add_header 'Access-Control-Allow-Origin' '*';` for `POST` and then hide it?

ansible/roles/mn-evo-services/templates/dapi-envoy.yaml.j2
```diff
@@ -20,12 +20,12 @@ static_resources:
                   cluster: echo_service
```
shumkov: As I understood it's only for Core and Platfrom services which are listening on 3005 port. We also have TransactionsFilterStream on 3006, so we need to add one more listener?

ansible/roles/mn-evo-services/templates/dapi-nginx.conf.j2
shuplenkov: It's hiding Access-Control-Allow-Origin received from envoy. If we don't hide - we'll get double header, even if the value is the same. And another thing that I don't know how to fix - envoy sends this header equals 'null' instead of '*'.

ansible/roles/mn-evo-services/templates/dapi-envoy.yaml.j2
shuplenkov: yes I think

shumkov: It seems like we need just to add one more virtual host and cluster for TransactionsFilterStream

ansible/roles/mn-evo-services/templates/dapi-nginx.conf.j2
shumkov: In this case, probably we just need to remove cors configuration from envoy?

shuplenkov: Yep, if we don't want to use envoy directly.

terraform/aws/security_groups.tf
```diff
@@ -249,6 +249,18 @@ resource "aws_security_group" "masternode" {
     ]
   }
 
+  # Envoy
+  ingress {
```
shumkov: Do we need to open Envoy for wild?

ansible/roles/mn-evo-services/templates/dapi-envoy.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+              routes:
+              - match: { prefix: "/" }
+                route:
+                  cluster: echo_service
+                  max_grpc_timeout: 0s
+          http_filters:
+          - name: envoy.grpc_web
+          - name: envoy.router
+  clusters:
+  - name: echo_service
```
shumkov: Let's rename it to `core_and_platform`?

ansible/roles/mn-evo-services/templates/dapi-envoy.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+    filter_chains:
+    - filters:
+      - name: envoy.http_connection_manager
+        config:
+          codec_type: auto
+          stat_prefix: ingress_http
+          route_config:
+            name: local_route
+            virtual_hosts:
+            - name: local_service
```
shumkov: Let's rename it to `core_and_platform`

PR #135, opened 2020-03-18
feat: introduce monitoring alarms

v2nek:


ansible/deploy.yml
```diff
@@ -2,6 +2,16 @@
 
 # Bootstrap
 
+- hosts: all
```
shumkov: We have it already below

terraform/aws/monitoring.tf
```diff
@@ -0,0 +1,104 @@
...
+  threshold           = "80"
+
+  dimensions = {
+    InstanceId = local.instance_ids[count.index]
+    MountPath  = "/"
+    Filesystem = "/dev/nvme0n1p1"
+  }
+
+  alarm_description = "This metric monitors ec2 disk utilization"
+  #  alarm_actions     = ["${aws_autoscaling_policy.bat.arn}"]
```
shumkov: Do we need it?

v2nek: this is required for notification actions (for example push message to slack, so yes, it will be used on later stages)

PR #136, opened 2020-03-19
fix: facts gathering

v2nek:


terraform/aws/iam.tf
```diff
@@ -0,0 +1,38 @@
+resource "aws_iam_role" "monitoring" {
```
shumkov: I guess we have all these changes in #135

PR #137, opened 2020-03-19
chore: add sleep before starting Drive API

jawid-h:


shumkov:
@jawid-h It looks more like a fix?

PR #139, opened 2020-03-19
feat: nice dns names

v2nek:


terraform/aws/main.tf
```diff
@@ -78,8 +78,16 @@ resource "aws_subnet" "public" {
   }
 }
 
+locals {
```
shumkov: We have already devnet name in our CLI, so let's just pass it:
- https://github.com/dashevo/dash-network-deploy/blob/3f23a9f400ece89c4aee996a231f9f98a2cfb836/lib/cli/init.sh#L7
- https://github.com/dashevo/dash-network-deploy/blob/3f23a9f400ece89c4aee996a231f9f98a2cfb836/lib/cli/terraform.sh#L49

terraform/aws/variables.tf
```diff
@@ -104,3 +104,7 @@ variable "vpn_enabled" {
   description = "setup instance for vpn"
 }
 
+variable "main_domain" {
```
shumkov: I wouldn't set the default value since the community will use this tool in order to spin up own devnets.

terraform/aws/main.tf
v2nek: This hack is to erase "devnet-" from network name

terraform/aws/variables.tf
v2nek: this still can be re-defined via tfvars file.

shumkov: Yes, but we should force people redefine it. And I would do that optional, maybe not everybody need it.

terraform/aws/main.tf
shumkov: Yes, but you can use env variable from cli

v2nek: with cli we are still used prefixed name, so we can rely on using this variable.

terraform/aws/variables.tf
v2nek: this one fixed

terraform/aws/main.tf
shumkov: `NETWORK_DEVNET_NAME` is what you need.

v2nek: done

PR #179, opened 2020-08-17
fix: increase disk space for masternode root volume

v2nek:


PastaPastaPasta:
What happens if this value is surpassed? Will the node not be able to keep up?

shumkov:
Nope, terraform will destroy and then recreate the node

PR #199, opened 2020-12-18
feat: validate .env in init

strophy:
Checking for presence of the `.env` and loading its content was previously done in two places: in the `dash-network` command wrapper and in the `generate` command. This PR removes the duplicate code and moves the check into the init function.

This fixes the chicken/egg problem of not having an `.env` file and not being able to create one because `init.sh` checks for its existence first, and also the problem of overwriting existing `.env` files when running `generate`.

## What was done?
Remove `.env` forced exit in `dash-network` and check/generate in `init.sh` instead of in the generate command.

## How Has This Been Tested?
Tested running `bin/generate` from shell and `dash-network generate` with docker.

## Breaking Changes
None

## Checklist:

### DIFF
--- a/bin/dash-network
+++ b/bin/dash-network
@@ -9,6 +9,12 @@ if ! [[ -x "$(command -v docker)" ]]; then
   exit 1
 fi
 
+# Load configuration	
+
+if [[ -f ".env" ]]; then
+    source ".env"
+fi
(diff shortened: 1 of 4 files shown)


shumkov (approved):
Good job! 👍

PR #231, opened 2021-02-23
feat: make logs instance configurable

strophy:
This PR allows configuration of the instance size of the logs node.

Logs node was too slow and not configurable.

## What was done?
Specify defaults and add variables to configure AWS instance size and JVM heap size.

## How Has This Been Tested?
Tested on testnet

## Breaking Changes
None

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -125,3 +125,4 @@ mn_evo_services_path: '{{ dashd_home }}/{{ mn_evo_services_compose_project_name 
 elastic_username: elastic
 elastic_password: 
 kibana_encryptionkey: 
+elastic_max_heap_size: 
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -125,4 +125,4 @@ mn_evo_services_path: '{{ dashd_home }}/{{ mn_evo_services_compose_project_name 
 elastic_username: elastic
 elastic_password: 
 kibana_encryptionkey: 
-elastic_max_heap_size: 
+elastic_max_heap_size: 2g
--- a/terraform/aws/variables.tf
+++ b/terraform/aws/variables.tf
@@ -153,3 +153,8 @@ variable "logs_node_disk_size" {
   description = "Disk size of log nodes"
   default     = 100
 }
+
+variable "logs_node_instance_type" {
+  description = "Instance type of log nodes"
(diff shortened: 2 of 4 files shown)


ansible/group_vars/all
```diff
@@ -125,3 +125,4 @@ mn_evo_services_path: '{{ dashd_home }}/{{ mn_evo_services_compose_project_name
 elastic_username: elastic
 elastic_password: 
 kibana_encryptionkey: 
+elastic_max_heap_size:
```
shumkov, line 128: What is the minimal recommended heap size? I would keep something default here.

terraform/aws/variables.tf
```diff
@@ -153,3 +153,8 @@ variable "logs_node_disk_size" {
   description = "Disk size of log nodes"
   default     = 100
 }
+
+variable "logs_node_instance_type" {
```
shumkov, line 157: It seems it highly depends on size of the network, but elastc recommends 2xlarge as minimal setup. Probably we should set xlarge at least?

shumkov (changes requested):

ansible/group_vars/all
strophy: Default is 512MB, recommended is 50% of system memory. We already have to define the above variables before deploy is possible, so maybe just add a comment here about requirements and let the user define it depending on the node instance they select?

terraform/aws/variables.tf
strophy: I think this is something we will have to learn over time. For smaller networks with ~12 nodes (theoretically this will be the primary use of the tool, right?) then I think t3.medium would be ideal. For testnet, we obviously need something larger, depending on what we are logging.

ansible/group_vars/all
shumkov: We have default instance for logs so we know default memory size, right? Why not just put it here. If you change the instance you need fo change heap size as well.

terraform/aws/variables.tf
shumkov: Yeah. It would be nice to have in future automatic adjustments based on number of nodes. So you set only how many masternodes you want and the tool generate approprate configs for terraform and ansible.

shumkov (approved):
👍

PR #238, opened 2021-05-26
feat: integrate tenderdash 0.4.x

strophy:
## Issue being fixed or feature implemented

Tenderdash 0.4.0 has been released and must be supported in the deploy tool to deploy to Dash networks.

## What was done?

Added vars for new image versions

## How Has This Been Tested?

Tested on Palinka

## Breaking Changes

Will not work with Tenderdash < 0.4.0

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #239, opened 2021-06-14
feat: pre-generate tenderdash Ed25519 keys

strophy:

Tenderdash setup is unnecessarily slow and complex due to the need to run `tenderdash init`. Since validator config is now dynamic we can generate tenderdash Ed25519 keys in JavaScript and template all necessary files for tenderdash startup without the init tasks.

## What was done?

- [x] Generate and manipulate keys into correct format
- [ ]  Replace tenderdash init with templated files

## How Has This Been Tested?

- [ ] How should I test this? By making small test networks on AWS?

## Checklist:


bin/generate
```diff
@@ -6,7 +6,7 @@ set -ea
 
 CMD_USAGE="Generate network configs
 
-Usage: generate <network_name> <masternodes_count>"
+Usage: generate <network_name> <masternodes_count> <seeds_count>"
```
shumkov, line 9: What if we change seed count from argument to option because in 99% you don't want to change and default value (2) is good enough.

lib/configGenerator/generateAnsibleConfig.js
```diff
@@ -28,6 +28,49 @@ async function generateAnsibleConfig(network, networkName, masternodesCount) {
...
+    // Strip static metadata from keys so they are 32 bytes and
+    // concatenate to 64 byte Tenderdash cached key
+    // Is this safe in memory or do I need to consider these cases?
+    // https://stackoverflow.com/a/54646864/1661847
+    // https://nodejs.org/api/buffer.html#buffer_buf_byteoffset
+    const cachedKey = Buffer.from([...privateKey.slice(16), ...publicKey.slice(12)]);
+
+    // Derive ID from key
+    const publicKeyHash = crypto.createHash('sha256')
+      .update(Buffer.from([...publicKey.slice(12)]))
```
shumkov, line 48: Buffer#slice returns buffer so you don't need Buffer#from here

lib/configGenerator/generateAnsibleConfig.js
```diff
@@ -28,6 +28,49 @@ async function generateAnsibleConfig(network, networkName, masternodesCount) {
...
+      privateKeyEncoding: { format: 'der', type: 'pkcs8' },
+      publicKeyEncoding: { format: 'der', type: 'spki' },
+    });
+
+    // Strip static metadata from keys so they are 32 bytes and
+    // concatenate to 64 byte Tenderdash cached key
+    // Is this safe in memory or do I need to consider these cases?
+    // https://stackoverflow.com/a/54646864/1661847
+    // https://nodejs.org/api/buffer.html#buffer_buf_byteoffset
+    const cachedKey = Buffer.from([...privateKey.slice(16), ...publicKey.slice(12)]);
```
shumkov, line 44: You can use Buffer.concat

shumkov (changes requested):

bin/generate
strophy: Does this look right? I tried to put the options before the arguments but I can't understand the bash syntax, even like this I'm 100% sure if `for i in "${@:3}"` is correct - I can't break it, but I also can't figure out what it actually does.

lib/configGenerator/generateTerraformConfig.js
```diff
@@ -5,6 +5,7 @@ async function generateTerraformConfig(network, networkName, masternodesCount) {
     node_count: 1,
     miner_count: 1,
     masternode_count: masternodesCount,
+    main_domain: '"networks.dash.org"',
```
shumkov, line 8: I guess we shouldn't set dash domain as default value since we expect other organizations and devs use it

ansible/roles/mn-evo-services/tasks/tendermint.yml
```diff
@@ -9,29 +9,37 @@
...
+  file:
+    path: '{{ mn_evo_services_path }}/tendermint/data'
+    state: directory
+    owner: '{{ tendermint_uid }}'
+    group: '{{ tendermint_gid }}'
+    mode: 0750
 
-- set_fact:
-    tendermint_node_id: '{{ tendermint_node_id.stdout }}'
+- name: create Tendermint node_key.json
```
shumkov, line 29: you can use `loop` to template many files https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html

ansible/roles/mn-evo-services/tasks/tendermint.yml
```diff
@@ -9,29 +9,37 @@
...
 
-- set_fact:
-    tendermint_validator_key: '{{ tendermint_validator_key.stdout }}'
+- name: create Tendermint priv_validator_key.json
+  template:
+    src: 'roles/mn-evo-services/templates/tendermint/priv_validator_key.json.j2'
+    dest: '{{ mn_evo_services_path }}/tendermint/config/priv_validator_key.json'
+    owner: '{{ tendermint_uid }}'
+    group: '{{ tendermint_gid }}' 
+  when: skip_validator is not defined or skip_validator == false
```
shumkov, line 42: I guess you can put these files for all nodes like `tendermint init` did previously. It doesn't matter.

ansible/deploy.yml
```diff
@@ -217,6 +216,8 @@
     - role: activate-dashd-sporks
       when: dash_network != "mainnet" and dash_network != "testnet"
 
+# Wait for chainlocks here, then restart platform services?
```
shumkov, line 219: Why?

ansible/deploy.yml
```diff
@@ -97,11 +97,38 @@
...
+  pre_tasks:
+    - set_fact:
+        node: "{{ seed_nodes[inventory_hostname] }}"
+      tags: always
+      when: inventory_hostname in seed_nodes
+    - fail:
+        msg: Seed node not defined in tenderdash config
+      when: node is not defined
+
+    - name: Check if genesis file exists
```
shumkov, line 113: Why do we still need genesis time?

shumkov (changes requested):

ansible/deploy.yml
shumkov: Agreed to generate genesis time in deploy config generator

ansible/deploy.yml
```diff
@@ -111,33 +138,24 @@
         - dashd
     - role: mn-evo-services
       vars:
-        skip_tenderdash_genesis: true
-        skip_services_start: true
+        skip_validator: true
```
shumkov, line 141: You don't need it anymore

strophy:
Closed in favor of #238

PR #255, opened 2021-08-31
feat: atomic latency option added to Ansible

ktechmidas:

To add "random" latency at the network level to the masternode servers through Ansible in order to test Platform under latency more similar to how mainnet operates. This will apply to devnet/testnet only

## What was done?
* Two roles have been added to ansible as the underlying Linux command is not atomic
  * The first is to add network latency if latency == 1 in the ansible vars
  * The second will set the latency to nothing if latency variable is 0 (This is so each command can run atomically)

This is achieved using qdiscs (Linux traffic control) and applies to all traffic to and from the masternode in question. Once you set latency to 1 it will apply to all masternodes

Note that right now we are simply using a shell command to achieve this as there is no underlying Ansible module which can add latency. This will be revisited later with an actual module...

## How Has This Been Tested?

Tested by bringing up my own devnet and applying the modules then attempting a simple ping.

* With latency = 0 (default)

ubuntu@masternode-4:~$ ping google.com
PING google.com (142.250.217.78) 56(84) bytes of data.
64 bytes from sea09s29-in-f14.1e100.net (142.250.217.78): icmp_seq=1 ttl=90 time=8.54 ms
64 bytes from sea09s29-in-f14.1e100.net (142.250.217.78): icmp_seq=2 ttl=90 time=8.53 ms

* With latency = 1

buntu@masternode-4:~$ ping google.com
PING google.com (142.250.217.78) 56(84) bytes of data.
64 bytes from sea09s29-in-f14.1e100.net (142.250.217.78): icmp_seq=1 ttl=90 time=108 ms
64 bytes from sea09s29-in-f14.1e100.net (142.250.217.78): icmp_seq=2 ttl=90 time=102 ms

## Breaking Changes

None, since it's off by default, with latency = 1 I haven't noticed any breaking changes yet on my devnet.

It will add an "empty" qdisc when no latency is applied

## Checklist:

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -5,220 +5,221 @@
 # Remove latency on masternodes before deployment starts (if enabled)
 - hosts: masternodes
   become: true
-    - role: remove-fake-latency
+  roles: 
+    -  remove-fake-latency
 
 
-- hosts: all
-  gather_facts: true
-  become: true
-  pre_tasks:
-    - name: check if inside AWS
-      uri:
-        url: http://169.254.169.254/latest/meta-data
-        timeout: 2
-      register: aws_uri_check
-      failed_when: False
-    - set_fact:
-        is_aws_environment: "{{ aws_uri_check.status == 200 }}"
-  roles:
-    - role: aws
-      when: is_aws_environment
-    - swap
+# - hosts: all
+#   gather_facts: true
+#   become: true
+#   pre_tasks:
+#     - name: check if inside AWS
+#       uri:
+#         url: http://169.254.169.254/latest/meta-data
+#         timeout: 2
+#       register: aws_uri_check
+#       failed_when: False
+#     - set_fact:
+#         is_aws_environment: "{{ aws_uri_check.status == 200 }}"
+#   roles:
+#     - role: aws
+#       when: is_aws_environment
+#     - swap
 
-# Setup VPN
+# # Setup VPN
 
-- hosts: vpn
-  become: true
-  roles:
-    - role: openvpn
-      when: openvpn_enabled
-      tags: vpn
+# - hosts: vpn
+#   become: true
+#   roles:
+#     - role: openvpn
+#       when: openvpn_enabled
+#       tags: vpn
 
-# Setup Python and Docker
+# # Setup Python and Docker
 
-- hosts: all
-  become: true
-  tasks:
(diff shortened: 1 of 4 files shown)


shumkov:
I wouldn't consider it as a breaking change since it's a new functionality so if you don't touch it then everything works the same as before.

ansible/deploy.yml
```diff
@@ -214,6 +214,16 @@
       when: evo_services
     - role: elastic-beats
 
+
+# Set up fake latency on masternodes (if enabled)
+- hosts: masternodes
+  become: true
+  roles:
+    - role: add-fake-latency
+      when: latency == 1
```
shumkov, line 223: It seems for other variables in the project we use booleans.

ansible/group_vars/all
```diff
@@ -139,3 +139,9 @@ elastic_username: elastic
 elastic_password:
 kibana_encryptionkey:
 elastic_max_heap_size: 2g
+
+#Fake latency stuff
+
+latency: 0
```
shumkov, line 145: What if call it more specific like `masternodes_network_latency`?

ansible/group_vars/all
```diff
@@ -139,3 +139,9 @@ elastic_username: elastic
 elastic_password:
 kibana_encryptionkey:
 elastic_max_heap_size: 2g
+
+#Fake latency stuff
+
+masternode_network_latency: false
+latency_min: 10
+latency_max: 100
```
shumkov, line 147:
```suggestion
masternode_network_latency_min: 10
masternode_network_latency_max: 100
```
latency_max: 100

shumkov:
As we discussed the latency thing affects deployment speed too. Is it possible to disable latency before we do deploy and enable it back as the last step?

ansible/deploy.yml
```diff
@@ -2,217 +2,234 @@
 
 # Bootstrap
 
-- hosts: all
```
shumkov, line 5: It seems you commited something wrong

shumkov (changes requested):

ansible/deploy.yml
ktechmidas: Correct, I was testing... fixed it :|

shumkov (approved):
LGTM 👍

PR #289, opened 2022-04-12
fix: the input device is not a TTY

strophy:

Running smoke tests in CI from docker wrapped in the dash-network command resulted in the error `the input device is not a TTY`. It is not necessary to interact with the terminal when running smoke tests, so we should not keep STDIN open without being attached.

## What was done?

Remove interactive argument passed to `docker run`

## How Has This Been Tested?

In CI

## Breaking Changes

None

## Checklist:

### DIFF
--- a/bin/dash-network
+++ b/bin/dash-network
@@ -35,7 +35,7 @@ fi
 
 # Run container with specified command
 
-docker run -ti --rm \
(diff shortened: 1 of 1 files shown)


shumkov (approved):
👍

PR #292, opened 2022-04-29
Fix: 5 minute pause

ktechmidas:
A 5 minute pause is needed here for larger devnets *if* minimumblocks is above 4032 (which it will be for QR for future devnets)

The reason for this is the blocks are a lot slower (2.5m) for each subsequent block at this point so the protx's don't propagate properly until more blocks have passed. I suspect the larger the devnet, the more of an issue this is going to be, the 5 minute pause should work on most current devnets.

Unfortunately it skips the "wait 1 block" routine

A new PR will come eventually to fix this properly

Example: 
image

5 minute pause until we figure out what's causing some protx's to not be in the first block. 

## What was done?
5 minute pause

## How Has This Been Tested?
Deployed Malort

## Breaking Changes
None

## Checklist:

- [x ] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have added or updated relevant unit/integration/functional/e2e tests
- [ x] I have made corresponding changes to the documentation

### DIFF
--- a/ansible/roles/mn-createprotx/tasks/main.yml
+++ b/ansible/roles/mn-createprotx/tasks/main.yml
@@ -40,6 +40,10 @@
     num_blocks: 1
     balance_needed: 0
 
+- name: Pause for 5 minutes to allow propogation of protx's
+  pause:
+    minutes: 5
+
 # verify
 
 - name: get list of ProTx transactions from the wallet


ansible/roles/mn-createprotx/tasks/main.yml
```diff
@@ -40,6 +40,10 @@
     num_blocks: 1
     balance_needed: 0
 
+- name: Pause for 5 minutes to allow propogation of protx's
+  pause:
+    minutes: 5
```
shumkov, line 45: Should we add a condition and pause it only for "larger devnets if minimumblocks is above 4032"?

ktechmidas: We can, but I'm hoping to get rid of this in the next few weeks and replace it with proper code to check that all protx's have gone through before it starts the next part.

This is just a quick bandaid which me and strophy can comment in and out depending on what's needed over the next few weeks

PR #293, opened 2022-05-10
feat: write protx values to config inventory file

strophy:

Running test suite against an arbitrary network in CI requires specification of the protx values in the new `MASTERNODE_REWARD_SHARES_OWNER_PRO_REG_TX_HASH` variable. We therefore need to add protx values for the registered masternodes to the inventory after they have been registered on the network, half way through the deploy process.

## What was done?

- Fix warnings about deprecated Ansible `include` statement in all files
- Add a new task in the `mn-init` role to write the protx values to the inventory file when known

## How Has This Been Tested?

Locally against valhalla and testnet

## Breaking Changes

Preview output is here: https://github.com/dashevo/dash-network-configs/pull/66

## Checklist:

### DIFF
--- a/ansible/roles/generate-blocks/tasks/main.yml
+++ b/ansible/roles/generate-blocks/tasks/main.yml
@@ -8,8 +8,8 @@
       - regtest
       - devnet
 
-- include: regtest_generate.yml
+- include_tasks: regtest_generate.yml
   when: dash_network in generate_networks
 
-- include: wait_for_blocks.yml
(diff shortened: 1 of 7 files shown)


ktechmidas (dismissed):

shumkov (commented):
I feel like updating the existing config which is supposed to be updated manually will lead to some issues. At least you will lose formatting and comments. Maybe, it's better to update inventory file: it's generated and we are not touching it manually. Plus protx hash var will be available in absible tasks.

PR #303, opened 2022-06-17
perf: speedup core smoke test

strophy:

Smoke tests were timing out without returning errors

## What was done?

- Move collect blockchaininfo call from before function to separate test
- Refactor blockchaininfo call to timeout on individual node failure

## How Has This Been Tested?

Locally and in [CI running against test build here](https://github.com/dashevo/dash-network-ci/runs/6922072178?check_suite_focus=true)

## Breaking Changes

None

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@dashevo/dapi-client": "^0.22.0",
         "@dashevo/dashcore-lib": "~0.19.30",
-        "@dashevo/dashd-rpc": "^2.3.1",
+        "@dashevo/dashd-rpc": "dashevo/dashd-rpc#timeout",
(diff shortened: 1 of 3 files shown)


strophy:
@shumkov I've set `dashevo/dashd-rpc#timeout` as a dependency so we can see the `dashd-rpc` timeout feature changes from https://github.com/dashevo/dashd-rpc/pull/45 and built a test image as `strophy/dash-network-deploy:0.22.1-beta1`. You can see this image running in CI here: https://github.com/dashevo/dash-network-ci/actions

The first batch of tests progresses much faster now, but I wonder if we should rewrite the masternode list check to either use the same design with promises, or query single nodes only until one of them responds, then use that response to check all nodes. What do you think?

shumkov (approved):
Good job! 👍

PR #309, opened 2022-07-04
chore: Remove dashd_evo_image (use dashd_image)

nmarley:
This consolidates the 2 variables `dashd_evo_image` and `dashd_image`. There is no need for both of them, since they are both refer to the same service.

This is catching me out on some changes I'm making on master even through it's been merged in `v0.23-dev`,  so I'm opening this as a hotfix PR into master as well. There should be no reason this can't be merged for testnet, not sure about `devnet-333`.


shumkov (approved):
👍

PR #310, opened 2022-07-04
fix: Use getgovernanceinfo in RPC check

nmarley:

The `getblockchaininfo` call is heavy, consumes a lot of resources. Since we are only using this to check if RPC is ready, we can use something light like `getgovernanceinfo` instead. This saves 2-3 seconds each time it's called.

## What was done?

Changed `getblockchaininfo` to `getgovernanceinfo` in dashd RPC check.

## How Has This Been Tested?

I tested this by running the dashd play on a testnet masternode, once.

## Breaking Changes

n/a

## Checklist:


shumkov (approved):
👍

PR #313, opened 2022-07-04
chore: Create dashcore dir for each user in a loop

nmarley:

User dashcore directories are not consistent among the 3 users that we have on MNs.

## What was done?

Adds a loop to create .dashcore directory for root, dash and ubuntu users.

## How Has This Been Tested?

Tested on some testnet masternodes using dashd role.

## Breaking Changes

n/a

## Checklist:

### DIFF
--- a/ansible/group_vars/all
+++ b/ansible/group_vars/all
@@ -64,6 +64,17 @@ insight_port: 3001
 # DashCore Docker image
 dashd_image: dashpay/dashd
 
+users_to_manage:
+- user: "ubuntu"
+  group: "ubuntu"
+  dir: "/home/ubuntu"
+- user: "root"
+  group: "root"
+  dir: "/root"
+- user: "dash"
+  group: "dash"
+  dir: "/dash"
+
 dashd_user: dash
 dashd_group: dash
 dashd_home: /dash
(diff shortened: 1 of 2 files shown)


nmarley:
PR is into master so we can use this on testnet

ansible/group_vars/all
```diff
@@ -64,6 +64,17 @@ insight_port: 3001
...
+users_to_manage:
+- user: "ubuntu"
+  group: "ubuntu"
+  dir: "/home/ubuntu"
+- user: "root"
+  group: "root"
+  dir: "/root"
+- user: "dash"
+  group: "dash"
+  dir: "/dash"
```
strophy, line 76: Let's specify the dash user using vars like `{{ dashd_home }}` so we don't have to define them twice (also on line 78-80).

nmarley: Yes, in theory. I do plan to remove the dash user in the future if there is no solid justification for keeping it,  but this can keep it simpler for now.

shumkov:
Just wondering, why do we need three users for dash core?

strophy (approved):
:+1:

strophy:
> Just wondering, why do we need three users for dash core?

We don't, I guess it is just for legacy reasons as the tool was being developed. We plan to tidy up and remove the `dash` user in the future.

PR #314, opened 2022-07-04
chore: Consolidate dash.conf in one place and symlink

nmarley:

The dash.conf file is copied around for different users for dash-cli ease of use, but it's easy to get divergent (differences in the file).

## What was done?

This PR adds the file in one single location (`/etc/dash.conf`) and then symlinks it into user directories so that dash-cli can be called easier by any of the users.

Needs #313 first.

## How Has This Been Tested?

Tested on some testnet masternodes using dashd role.

## Breaking Changes

n/a

## Checklist:


nmarley:
PR is into master so we can use this on testnet

shumkov (approved):

PR #315, opened 2022-07-04
feat: add powtargetspacing to dash.conf vars

strophy:

Dash Core now supports the `powtargetspacing` setting

## What was done?

- Add powtargetspacing

## How Has This Been Tested?

On `devnet-jack-daniels`

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/roles/dashd/templates/dash.conf.j2
+++ b/ansible/roles/dashd/templates/dash.conf.j2
@@ -90,6 +90,9 @@ llmqinstantsenddip0024=llmq_devnet
 minimumdifficultyblocks={{ dashd_minimumdifficultyblocks }}
 highsubsidyblocks=500
 highsubsidyfactor=10
+{% if dashd_powtargetspacing is defined %}
+powtargetspacing={{ dashd_powtargetspacing }}
+{% endif %}
 {% endif %}
 
 {% if dashd_debug == 1 %}


ktechmidas:
LGTM

ktechmidas (approved):

ktechmidas:
(although, do we define this manually in the config file? Nothing in bin/generate for this yet?)

ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -90,6 +90,9 @@ llmqinstantsenddip0024=llmq_devnet
 minimumdifficultyblocks={{ dashd_minimumdifficultyblocks }}
 highsubsidyblocks=500
 highsubsidyfactor=10
+{% if dashd_powtargetspacing is defined %}
```
nmarley, line 93: @strophy Is this only for devnets?

strophy: @PastaPastaPasta @ogabrielides ^^

strophy:
> (although, do we define this manually in the config file? Nothing in bin/generate for this yet?)

I didn't add this to the config generator because there is no good default value. For most "normal" devnets this should not be defined, so I think the correct approach is to conditionally include it in config only if defined manually.

ansible/roles/dashd/templates/dash.conf.j2
ogabrielides: Correct

nmarley (approved):
utACK

PR #320, opened 2022-07-06
feat: Write and use docker-compose file for dashcore

nmarley:

Currently dashcore must be managed only via the Ansible play. This will allow management of the dashcore container locally,  and give the ability to locally tweak the config (image, command, etc) via the docker-compose.yml file. This is (I believe) the only service we run that does not use docker-compose in the tool.

## What was done?

Adds a docker-compose.yml file for dashcore, and uses docker-compose to manage the container. Also removes the play which users docker container to manage it.

## How Has This Been Tested?

Tested on masternode on testnet.

## Breaking Changes

n/a

## Checklist:


nmarley:
Note: This has also been tested and validated on testnet seed nodes now.

strophy (approved):
utACK

PR #325, opened 2022-07-14
feat: conditional core/platform elastic beats

strophy:

During recent core testing, it was common to recreate the core container without replaying the remainder of the playbook. Because `elastic-beats` was only set to run for platform deployment, this means that core logs were no longer captured if the platform deployment did not run.

## What was done?

This PR adds `elastic-beats` to core plays on all nodes, including miner, wallet and web nodes for the first time. Running core plays will now also update filebeat with new container IDs, and platform config is conditionally deployed on each run of the play, depending on whether platform is deployed or not.

## How Has This Been Tested?

- On devnet-jack-daniels

## Breaking Changes

@nmarley is this still idempotent? I'm not quite sure how to reason about deploying a single config to cover all three services at different times in the playbook.

## Checklist:

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -98,6 +98,7 @@
       dashd_zmq: true
       tags:
         - dashd
+    - role: elastic-beats
 
 - hosts: miners
   become: true
@@ -105,6 +106,7 @@
     - role: dashd
       tags:
         - dashd
+    - role: elastic-beats
 
 # Setup core on seed nodes
 
@@ -117,6 +119,7 @@
       dashd_indexes: true
       tags:
         - dashd
+    - role: elastic-beats
 
 # Setup core on masternodes
 
@@ -138,6 +141,7 @@
       tags:
         - dashd
     - mn-sentinel
+    - elastic-beats
 
 # Generate first block on seed node
 # so that all nodes leave IBD mode (required for mining)
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -98,7 +98,7 @@
       dashd_zmq: true
       tags:
         - dashd
-    - role: elastic-beats
+    - elastic-beats
 
 - hosts: miners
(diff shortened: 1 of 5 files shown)


nmarley:
Thanks @strophy . I'm not sure it's necessary to break these all out. Instead, can't we install them all at once on all nodes (not in core _or_ in platform, but before any of that gets installed, unconditionally).

If the file or file pattern/glob exists, then it should read it and send it up to elastic. If not then it will keep monitoring regulary for any files that pop up that do match that pattern, is my understanding.

https://www.elastic.co/guide/en/beats/filebeat/current/how-filebeat-works.html#_how_does_filebeat_keep_the_state_of_files


It doesn't seem right to install this w/core, for example, since I might want to install core w/o mucking about w/filebeats. Seems messier to me and I feel like this should perhaps just be 1-time install for all filebeats at server setup instead.

strophy:
Yes, I totally see your point about wanting to do this once only. But the entire purpose of this commit is to solve the problem that a once-only run of this role breaks when the core container is recreated. Core logs are picked up from e.g. `/var/lib/docker/containers/fe1f608003884cb093ed9b4fe673fbd3cf6b5ac8868ce677dc42c57b48762460/fe1f608003884cb093ed9b4fe673fbd3cf6b5ac8868ce677dc42c57b48762460.log`, so it is necessary to update filebeat every time the container ID changes. It would be nice to use a wildcard glob here, but `/var/lib/docker/containers/` contains multiple directories which may also contain a file that matches `*.log`, but is not the file we actually want to monitor. Maybe there is a clever way we can do this using environment variables, so it is only necessary to restart the filebeat service without retemplating all the config files like I have done here?

nmarley (approved):
utACK

I prefer to do this differently and send up everything, but don't want to hold this up and don't have the capacity to look into changing this or setting up logstash at this time.

PR #326, opened 2022-07-14
feat: logrotate for core debug.log

strophy:

Core logs were filling up the miner node on devnet-jack-daniels until it ran out of disk space.

## What was done?

Add logrotate to core debug.log

## How Has This Been Tested?

Tested on devnet-jack-daniels, but still needs to run through completely on a fresh deploy of a network with the miner logs problem to verify it is working.

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/roles/dashd/tasks/main.yml
+++ b/ansible/roles/dashd/tasks/main.yml
@@ -56,6 +56,26 @@
     - docker-compose.yml
     - .env
 
+- name: Configure log rotation
+  include_role:
+    name: arillso.logrotate
+  vars:
+    logrotate_use_hourly_rotation: true
+    logrotate_applications:
+      - name: core-logs
+        definitions:
+          - logs:
+              - "{{ dashd_home }}/.dashcore/{{ 'devnet-' + dash_devnet_name if dash_network == 'devnet' else 'testnet3' }}/debug.log"
+            options:
+              - rotate 7
+              - daily
+              - maxsize 1G
+              - missingok
+              - notifempty
+              - copytruncate
+              - compress
+              - delaycompress
+
 - name: Start dash core
   docker_compose:
     project_src: '{{ dashd_compose_path }}'
--- a/ansible/roles/dashd/tasks/main.yml
+++ b/ansible/roles/dashd/tasks/main.yml
@@ -56,46 +56,22 @@
     - docker-compose.yml
     - .env
 
-- name: Configure log rotation
+- name: Configure core debug.log rotation
   include_role:
-    name: arillso.logrotate
+    name: logrotate
   vars:
-    logrotate_use_hourly_rotation: true
-    logrotate_applications:
+    logrotate_options:
       - name: core-logs
-        definitions:
-          - logs:
-              - "{{ dashd_home }}/.dashcore/{{ dash_network_name if dash_network == 'devnet' else 'testnet3' }}/debug.log"
-            options:
-              - rotate 7
-              - daily
-              - maxsize 1G
-              - missingok
-              - notifempty
-              - copytruncate
-              - compress
-              - delaycompress
-
-- name: Register systemd presence
-  stat:
-    path: /lib/systemd/system
-  register: systemd_dir
-
-- name: Ensure logrotate runs hourly under systemd timer
-  lineinfile:
-    path: /lib/systemd/system/logrotate.timer
-    regexp: '^OnCalendar=hourly'
-    insertafter: '^OnCalendar=daily'
-    line: OnCalendar=hourly
-  when: systemd_dir.stat.exists
-
-- name: Enable and restart logrotate systemd timer
-  systemd:
-    name: logrotate.timer
-    state: restarted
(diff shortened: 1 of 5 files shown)


ansible/roles/dashd/tasks/main.yml
```diff
@@ -56,6 +56,26 @@
...
+- name: Configure log rotation
+  include_role:
+    name: arillso.logrotate
+  vars:
+    logrotate_use_hourly_rotation: true
+    logrotate_applications:
+      - name: core-logs
+        definitions:
+          - logs:
+              - "{{ dashd_home }}/.dashcore/{{ 'devnet-' + dash_devnet_name if dash_network == 'devnet' else 'testnet3' }}/debug.log"
```
nmarley, line 68: Is there some reason to construct the network name instead of using the `dash_network_name` variable?

ansible/roles/dashd/tasks/main.yml
```diff
@@ -56,6 +56,26 @@
...
+            options:
+              - rotate 7
+              - daily
+              - maxsize 1G
```
nmarley, line 72: 100M should be more than enough. 7GB just for logs (1GB * 7 log files) is a lot.

ansible/roles/dashd/tasks/main.yml
strophy: Good point!

ansible/roles/dashd/tasks/main.yml
strophy: True, but core in debug mode can generate a lot of logs on the miner when a bug such as `bad-qc-not-allowed` is encountered. We filled ~15GB of free space on the node in under 3-4h, so I think 1GB is a reasonable threshold if logrotate runs hourly. The main point of this is just to stop the drive from running out of space, the logs we want to keep will be safely stored in elasticsearch (after #325 is merged).

strophy:
Refactored and tested on a temporary devnet with a problem that causes a lot of output in `debug.log`, after first run the state is:
```
root@miner-1:/dash/.dashcore/devnet-sambuca# ls debug* -lha
-rw------- 1 dash dash 2.7G Jul 21 11:43 debug.log
-rw------- 1 dash dash 3.2G Jul 21 11:00 debug.log.1
```
After second run:
```
root@miner-1:/dash/.dashcore/devnet-sambuca# ls debug* -lha
-rw------- 1 dash dash  25M Jul 21 12:02 debug.log
-rw------- 1 dash dash 3.8G Jul 21 12:01 debug.log.1
-rw------- 1 dash dash  38M Jul 21 11:00 debug.log.2.gz
```

nmarley:
Thanks for this. It's not a bad idea but replacing the entire logrotate role seems to be more work and complexity than is necessary. I probably didn't explain myself well, sorry about that.

(This PR seems to be doing 2 things now also:1) add core logrotate, and 2) 

Here is an alternative for

nmarley:
Thanks for this. It's not a bad idea but replacing the entire logrotate role seems to be more work and complexity than is necessary. I probably didn't explain myself well, sorry about that.

This PR seems to be doing 2 things now also:

1. add core logrotate, and
2. remove Ansible role and replace w/our own

Here is an [alternative for the core bit](https://github.com/dashevo/dash-network-deploy/compare/v0.23-dev...nmarley:dash-network-deploy:core-logrotate-alt?expand=1).

Here is an [alternative for the replacement for platform](https://github.com/dashevo/dash-network-deploy/compare/v0.23-dev...nmarley:dash-network-deploy:platform-logrotate-alt?expand=1).

I would expect those to be separate PRs.

Note that as-is, **this alternative platform branch will break/not work unless tenderdash starts to run as uid 1001 also**. I'm not sure why we are running as `tmuser` but this should really be 1001 if we are to sync up the user and not have issues like this. (I've done this on a testnet node once and it worked fine, we just have to make sure and update the uid/gid on all previously created files.)

Here is my [alternative for running logrotate hourly](https://github.com/dashevo/dash-network-deploy/compare/v0.23-dev...nmarley:dash-network-deploy:logrotate-hourly?expand=1), if you need that. (Another separate PR I think.)

I haven't tested any of these yet but these are to give you an idea of what I was thinking, instead of using some pre-packaged Ansible role.

ansible/roles/dashd/tasks/main.yml
kxcd: What you need is a file like `/etc/logrotate.d/dashd` 
```
/dash/.dashcore/testnet3/debug.log {
        rotate 9
        copytruncate
        compress
        notifempty
        missingok
        size 1G
}
```
I have tested this and it creates good logs without missing any, I prefer compressed logs, use `zcat` and `zgrep` to view and search them.

Next edit the file `/usr/lib/systemd/system/logrotate.timer` and change the word `daily` to `hourly`.
```
OnCalendar=hourly
```
Then either reboot (preferred) or daemon-reload.

strophy:
Closed in favour of #346 and #347

PR #330, opened 2022-07-26
fix: use local directory in case of multiple users

ktechmidas:
For elastic-beats and vpn roles it will use the generic /tmp directory, this is an issue as these files are then "locked" by that user. In the case of using a manager node with multiple users the other users will then get "permission denied" when they attempt to do a deployment

elastic-beats and vpn roles now use local directory in the user's home to generate certificates

## How Has This Been Tested?
Deployed a devnet

## Breaking Changes
If another user deploys over the top, it will regenerate a certificates. This shouldn't be too much of an issue AFAIK


strophy (approved):
utACK

PR #331, opened 2022-08-05
perf: speedup sentinel tests

strophy:

Sentinel tests were running sequentially, making it extremely slow on large networks.

## What was done?

Parallelize tests

## How Has This Been Tested?

This current cannot be tested because we are using dev sentinel images on both `devnet-jack-daniels` and `testnet`. These images have log output, but one of the tests checks for no log output in the last 5 minutes, and therefore fails.

I am also unable to run anything against testnet, possibly due to a VPN error. I recreated the testnet VPN node from scratch without any change in behaviour, and had a call with @shumkov to try and resolve this without success. @nmarley I now know more about why the VPN is needed, but my knowledge of networking is insufficient to resolve the problem, so I need your help please.

## Breaking Changes

None

## Checklist:

### DIFF
--- a/test/smoke/sentinel.js
+++ b/test/smoke/sentinel.js
@@ -14,16 +14,6 @@ describe('Sentinel', () => {
 
       const promises = [];
 
-      function requestLogsStream(docker, hostName) {
-        return docker.getContainer(listContainers[hostName][0].Id).logs({
-          stdout: true,
-          stderr: true,
-          follow: 0,
-          since: (Math.floor(new Date() / 1000) - 300), // logs from last 5 mins
-          timestamps: true,
-        });
-      }
-
       for (const hostName of inventory.masternodes.hosts) {
         const timeout = 15000; // set individual docker client timeout
         const docker = new Docker({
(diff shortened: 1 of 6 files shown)


nmarley (dismissed):
concept ACK, I really am not deep enough in the code to be able to review this properly. I will say that the Sentinel tests are generally fast (around 0.2s on my machine), so this is more of a CI workflow adjustment than anything to do w/Sentinel.

It's also strange that we would run Sentinel unit tests for this. Those are more for developers of Sentinel itself and doesn't make sense here as an integration test. Why not just remove them from here instead? Why were they added? @shumkov

test/smoke/sentinel.js
```diff
@@ -21,21 +28,54 @@ describe('Sentinel', () => {
           }),
         };
 
-        const containers = await docker.listContainers(options);
+        const requestListContainers = docker.listContainers(options)
+          .then((result) => {
+            listContainers[hostName] = result;
```
shumkov, line 33: You can do everything in one step and get container log right here

shumkov, line 33: You can do everything in one step and get container log right here

shumkov (changes requested):

test/smoke/sentinel.js
strophy: Can you explain a bit more what you mean? The promises can't be resolved at the same time because the second promise depends on data from the first promise existing. I tried adding the call to get the lists in another `.then()` statement, but I can't figure out how to do this so it all resolves as a single promise...

strophy: @shumkov I think I got it working in a single promise now?

shumkov (approved):
👍

PR #332, opened 2022-08-08
perf: speedup tenderdash tests

strophy:

Tenderdash tests were running sequentially, making it extremely slow on large networks.

## What was done?

Parallelize tests

## How Has This Been Tested?

On devnet-jack-daniels

## Breaking Changes

None

## Checklist:

### DIFF
--- a/test/smoke/tendermint.js
+++ b/test/smoke/tendermint.js
@@ -7,8 +7,6 @@ const { variables, inventory, network } = getNetworkConfig();
 describe('Tendermint', () => {
   const masternodeStatus = {};
   const masternodeStatusError = {};
-  const masternodeStatusNodeInfo = {};
-  const masternodeStatusNetworkName = {};
 
   const masternodeNetInfo = {};
   const masternodeNetInfoError = {};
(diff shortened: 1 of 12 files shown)


test/smoke/tendermint.js
```diff
@@ -5,116 +5,134 @@ const getNetworkConfig = require('../../lib/test/getNetworkConfig');
...
+            let networkName = `dash-${network.name}`;
+            if (network.type === 'devnet' && variables.tenderdash_chain_id !== undefined) {
+              networkName = `dash-${variables.tenderdash_chain_id}`;
+            }
+
+            masternodeStatusNetworkName[hostName] = networkName;
+          })
+          .catch((e) => {
+            // eslint-disable-next-line no-console
+            console.log(e);
```
shumkov, line 45: Won't it be better if you keep this error and pass to fail function below?

nmarley (commented):
concept ACK, not able to give a proper review ATM

test/smoke/tendermint.js
strophy: Yes, I agree! But I couldn't figure out how to do this, because we already collect errors on line 33. I only added this because without catching errors and doing something (like `console.log()`) in the `.catch()` statement, the entire smoke test fails if a single node fails. I can't figure out why, I guess the error is not included in the response if the node isn't listening? I don't really understand how networking errors work or which module is actually throwing the error here, where can I learn about this stuff? How can I check network responses and see the status of a promise in the debugger?

PR #333, opened 2022-08-10
perf: speedup drive tests

strophy:

Drive tests were running sequentially, making it extremely slow on large networks.

## What was done?

Parallelize tests

## How Has This Been Tested?

On testnet (and devnet-jack-daniels with alternate echo string `0a0a080a0668656c6c6f21`

## Breaking Changes

Won't work with Drive 0.23, separate PR necessary

## Checklist:

### DIFF
--- a/test/smoke/drive.js
+++ b/test/smoke/drive.js
@@ -2,25 +2,32 @@ const net = require('net');
 
 const getNetworkConfig = require('../../lib/test/getNetworkConfig');
 
-const { variables, inventory } = getNetworkConfig();
+const { inventory } = getNetworkConfig();
 
 async function sendEcho(ip) {
   const echoRequestBytes = Buffer.from('0e12050a03010203', 'hex');
+  console.log(echoRequestBytes.toString('hex'));
 
   return new Promise((resolve, reject) => {
     const client = net.connect(26658, ip);
 
     client.on('connect', () => {
+      console.log('in connect statement');
       client.write(echoRequestBytes);
     });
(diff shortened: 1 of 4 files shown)


test/smoke/drive.js
```diff
@@ -28,32 +28,37 @@ async function sendEcho(ip) {
...
-        }
+      return Promise.all(promises).catch(() => Promise.resolve());
+    });
 
-        // eslint-disable-next-line no-underscore-dangle
-        await sendEcho(inventory._meta.hostvars[hostName].public_ip);
+    for (const hostName of allHosts) {
+      describe(hostName, () => {
+        it('should listen for ABCI connection', () => {
+          expect(echoInfo[hostName]).to.be.true();
```
shumkov, line 59: Will we see the error in this case?

strophy: Yes, when drive fails to respond to the ping we get:
```
  1) Smoke
       Drive
         All nodes
           masternode-2
             should listen for ABCI connection:
     AssertionError: expected [Error: Can't connect to ABCI port: timeout.] to be true
      at Context.<anonymous> (test/smoke/drive.js:59:48)
```

Is that good enough? Or should I collect errors in a different object?

shumkov (approved):
👍

test/smoke/drive.js
shumkov: I would recommend using `expect.fail` as a more idiomatic way.

PR #335, opened 2022-08-10
fix: openvpn persists after SIGINT

strophy:

OpenVPN remains connected after interrupting a smoke test run with Ctrl + C. Subsequent runs reuse the same VPN connection, meaning it is not possible to run smoke tests on a different network without first running `sudo kill $(pidof openvpn)`.

## What was done?

Catch `SIGINT` and give OpenVPN time to disconnect and destroy the client.

## How Has This Been Tested?

On `devnet-jack-daniels`

## Breaking Changes

This changes the prior behaviour, but I consider it a fix of incorrect behaviour rather than a breaking change of behaviour

## Checklist:

### DIFF
--- a/lib/test/bootstrap.js
+++ b/lib/test/bootstrap.js
@@ -22,5 +22,9 @@ after(shutdownVpnClient);
 
 process.on('uncaughtException', console.error);
 process.on('unhandledRejection', console.error);
+process.on('SIGINT', () => {
+  shutdownVpnClient();
+  process.exit(1);
+});
 
 global.expect = expect;
--- a/lib/test/bootstrap.js
+++ b/lib/test/bootstrap.js
@@ -22,8 +22,8 @@ after(shutdownVpnClient);
 
 process.on('uncaughtException', console.error);
 process.on('unhandledRejection', console.error);
-process.on('SIGINT', () => {
-  shutdownVpnClient();
+process.on('SIGINT', async () => {
+  await shutdownVpnClient();
   process.exit(1);
 });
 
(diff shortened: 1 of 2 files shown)


strophy:
I think this is a weak solution, it should be possible to wait for an event. I tried wrapping it in a promise like this https://stackoverflow.com/a/70789108 but I still can't get it to wait. Help!

shumkov:
@thephez https://hackernoon.com/graceful-shutdown-in-nodejs-2f8f59d1c357 and please google "async graceful shutdown"

thephez:
> @thephez https://hackernoon.com/graceful-shutdown-in-nodejs-2f8f59d1c357 and please google "async graceful shutdown"

@shumkov Interesting, but I think the comment was intended for @strophy? :slightly_smiling_face:

strophy:
OK, I read most of those articles and the main difference seems to be putting `process.exit()` in an arrow function to run after `shutdownVpnClient()` instead of making the initial function call async and using `await`. The articles all relate to express, which I think is a web server? But this approach doesn't solve the actual problem inside the `shutdown()` function, the `vpnManager.destroy()` code is never reached because I can't figure out how to wait for an event in JS. This was my original question.

kxcd:
Be careful with the terminology and what you are trying to do.
`process.on('SIGINT', async () => {`
If you are sending SIGINT to a process do not be suprised that process is still in the process table after you run it!  The signal is an interrupt, similar to pressing CTRL+Z on the keyboard and not CTRL+C.  To end a process that is not already hung, you want to send the SIGTERM (15) signal for a graceful shutdown, otherwise SIGKILL (9) for immediate destruction.

shumkov:
@thephez my bad :)

shumkov (approved):
Good job! 👍

PR #349, opened 2022-09-07
feat: multifaucet promo codes

strophy:

`faucet.testnet.networks.dash.org` did not support promo codes

## What was done?

Enable promo codes

## How Has This Been Tested?

On testnet

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/roles/multifaucet/tasks/main.yml
+++ b/ansible/roles/multifaucet/tasks/main.yml
@@ -26,3 +26,13 @@
     project_src: '{{ dashd_home }}/multifaucet'
     state: present
     build: yes
+
+- name: create promo codes table
+  community.docker.docker_container_exec:
+    container: multifaucet_db_1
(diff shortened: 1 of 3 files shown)


ktechmidas (approved):
LGTM

nmarley (commented):
I _think_ we had decided several months ago on using only the other faucet and not this one (at least for testnet -- might be a good change for devnets). Are you planning to start using this one instead or alongside the other one we maintain?

strophy:
I remember that too! I think we discovered that faucet.testnet.networks.dash.org has a bot that repeatedly attempts to drain the faucet - was this the reason we went with the other one? We tried for almost an hour to update the testnet-faucet.dash.org through terraform without success, so the thinking was that this one is easier to maintain because we understand how to do it while you are away @nmarley. Long term I think I would prefer pointing the other one at this one so we only have to maintain one faucet.

ktechmidas (approved):
LGTM

PR #353, opened 2022-09-12
perf: speedup karma tests

strophy:

Karma tests were running sequentially, making it extremely slow on large networks.

## What was done?

Parallelize tests.

## How Has This Been Tested?

On `devnet-jack-daniels`

## Breaking Changes

Doesn't work with older versions of test tools, #352 must be merged first

## Checklist:


strophy:
@nmarley here's another example of a common git mistake I make. I needed to use code from another PR which hadn't been reviewed and merged to `v0.23-dev` yet, and I didn't want to wait, so I made `speedup-karma-tests` based on the `update-deps` branch. `update-deps` was then merged to `v0.23-dev`, and I tried to both rebase and merge `v0.23-dev` into `speedup-karma-tests`. It always resulted in conflicts, but I don't understand why because I don't think any more commits were made before I started work on `speedup-karma-tests`. In the end merge failed with `ff` but worked with `ort`, but now I have merges showing both the commits from `update-deps`, and `speedup-karma-tests` and then one more commit on the top to merge the branch in. And yet somehow only the correct changes are shown in the diff, so I think we can still just squash merge this? But what is the right way to do this workflow? Rebase?

shumkov (approved):

PR #365, opened 2022-10-13
chore: change spork activation command

ktechmidas:
Core team have a new method for enabling sporks

## What was done?
Added new method for enabling sporks

## How Has This Been Tested?
Tested in various devnets, now that all devnets are testnet is < 18.1.0 we can go ahead and merge this change

## Breaking Changes
Deploy tool will no longer work on dashd versions < 18.1.0

## Checklist:

- [x ] I have performed a self-review of my own code

### DIFF
--- a/ansible/roles/activate-dashd-sporks/tasks/main.yml
+++ b/ansible/roles/activate-dashd-sporks/tasks/main.yml
@@ -17,7 +17,7 @@
   register: active_sporks
 
 - name: activate non active sporks
-  shell: dash-cli spork {{item}} 0
+  shell: dash-cli sporkupdate {{item}} 0
   with_items:
     - SPORK_2_INSTANTSEND_ENABLED
     - SPORK_3_INSTANTSEND_BLOCK_FILTERING


ansible/roles/activate-dashd-sporks/tasks/main.yml
```diff
@@ -16,8 +16,22 @@
   shell: dash-cli spork active
   register: active_sporks
 
+# TODO: ensure JQ installed
+- name: get dashcore version number
```
shumkov, line 20: Does it make senes since we aren't going to deploy Core lower 18.1.0 with Platform 0.23?

nmarley: Isn't it supposed to be a community tool? Since current version is v18.0, I wanted to make sure everyone is able to run this and we don't leave out everyone else except DCG.

Do you not agree?

strophy: While I agree with @nmarley on the side of correctness and backwards compatibility here, I think we need to also recognise the reality that we currently have no other known users of this tool, so we can assume with reasonable safety that any eventual future users would also not want to run a version of core lower than 18.1.0. Additional logic like this negatively affects readability and the already notoriously slow speed of the tool.

PastaPastaPasta: FYI 18.1.0 is in the process of being built and will be released today or tomorrow

nmarley (approved):
LGTM 👍

strophy (approved):

ansible/roles/activate-dashd-sporks/tasks/main.yml
strophy: This tool is designed for testing pre-release versions, so by definition we need to make updates like this here _before_ we can release 18.1.0 if we are following proper procedure. The only reason it didn't happen this way is because the test network was deployed from a branch in order to not cause delays.

PR #380, opened 2022-10-24
Fix: Ensure wallet1 is using the faucet privkey even if web isn't deployed

ktechmidas:
Moved importing the privkey for the faucet to the main section of dashd so can make multifaucet optional. It makes no sense that the whole mining wallet that the mn-init role relies on is done in multifaucet of all roles

## What was done?
Import privkey into new wallet1 before continuing setup

## How Has This Been Tested?
devnet-infra-argh

## Breaking Changes
None


ansible/roles/dashd/tasks/main.yml
```diff
@@ -96,5 +96,9 @@
   retries: 5
   delay: 10
 
+- name: import faucet privkey
+  when: inventory_hostname in groups['faucet_wallet']
+  shell: dash-cli importprivkey {{ faucet_privkey }}
```
strophy, line 101: Are you sure this works before the wallet is created?

strophy:
Good catch! Why do we need to run it twice, before and after the wallet is created?

ansible/roles/dashd/tasks/main.yml
ktechmidas: Oh that's a good catch, I've moved it below the "create_wallet" section

ktechmidas: So it should work on both <18.2 and >18.2 now

strophy (approved):
:+1: good change

PR #390, opened 2022-11-14
feat: single wallet node for wallet and mno

strophy:

Two wallet nodes are currently created for each network. Dash Core 18 implements support for multiple wallets, which we can use to reduce the node count. Dash Core 19 will disable automatic wallet creation, so this is a good time to re-examine this logic anyway since it will break soon. 

## What was done?

Refactor to use two wallets on a single wallet node, with the mno wallet only loaded when running protx-related commands. This allows us to specify the default wallet as the faucet wallet, so we can simplify all calls between the nodes and completely remove the extremely annoying `faucet_rpc_args` variable, which cause Ansible to run commands on a node other than the one defined in the role `hosts` variable, leading to regular confusion. We can also remove hacky workarounds designed to wait for transactions to appear between the two wallet nodes, since this now all occurs on the same node, further improving performance of the `mn_init` role.

This new design currently only supports a single wallet node, but it would probably be fairly easy to modify to support multiple MNOs per network, each owning a subset of the network masternodes, and receiving block rewards based on random mining and the number of masternodes that they own. Is this useful?

## How Has This Been Tested?

On `devnet-wallet`

## Breaking Changes

Not yet clear

## Checklist:


ktechmidas (approved):
utACK, looks good to me but let's see what it breaks :)

PR #391, opened 2022-11-16
fix: inventory is updated twice

strophy:

A few missing pieces of docs/bugs needed cleaning up after merging #390 

## What was done?

- Clean up unnecessary/incorrect comments and debug code
- Reset a variable that may potentially already contain data

## How Has This Been Tested?

On `devnet-jack-daniels`

## Breaking Changes

None

## Checklist:

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -52,6 +52,7 @@ done
 
 echo "Deploying $NETWORK_NAME network...";
 
+# TODO: this fails without an error if not at the HEAD of a given branch
 BRANCH=`git symbolic-ref --short -q HEAD`
 if [[ $BRANCH == "devnet-"* ]]; then
     if [[ $BRANCH != $NETWORK_NAME ]] && [[ $FORCED_DEPLOYMENT != 1 ]]; then
--- a/ansible/roles/mn_protx_config/tasks/find_protx.yml
+++ b/ansible/roles/mn_protx_config/tasks/find_protx.yml
@@ -1,6 +1,6 @@
 ---
 
-# Requires "masternode" and "masternode_name" variables
+# Requires "masternode" variable
 
 - name: Get protx of registered masternodes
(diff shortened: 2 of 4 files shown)


ktechmidas (approved):
LGTM

bin/deploy
```diff
@@ -52,6 +52,7 @@ done
 
 echo "Deploying $NETWORK_NAME network...";
 
+# TODO: this fails without an error if not at the HEAD of a given branch
```
nmarley, line 55: @strophy How does this fail? From my testing it will set BRANCH to an empty string, but that looks to be a valid condition below, since an empty string won't match anything.

strophy: @nmarley it fails with exit code `1`. Try this to reproduce:
```
git checkout v0.23-dev
git symbolic-ref --short -q HEAD // returns v0.23-dev
git checkout 7d32585bf04bd9510367064038d0a7ac31b02765
git symbolic-ref --short -q HEAD // fails
```

nmarley: Ok, now I understand. The script fails because `-e` is set above to fail upon any error. If you want an error you can just remove the `-q`:
```
-q, --quiet
    Do not issue an error message if the <name> is not a symbolic ref but a detached HEAD;
    instead exit with non-zero status silently.

Comment (nmarley) [reply bin/deploy:55]: Ok, now I understand. The script fails because `-e` is set above to fail upon any error. If you want an error you can just remove the `-q` from the `git symbolic-ref` command:

```
-q, --quiet
    Do not issue an error message if the <name> is not a symbolic ref but a detached HEAD;
    instead exit with non-zero status silently.
```

Or we can discuss a different fix internally, but that should be better than silently failing I would think.

strophy: @nmarley just an error is fine for now: https://github.com/dashpay/dash-network-deploy/pull/393

PR #393, opened 2022-11-18
fix: deploy fails silently if current branch not at HEAD

strophy:

`bin/deploy` would fail with exit code 1 and no visible error if deploying from a commit below HEAD on any given branch

## What was done?

Remove `-q` so the failing command displays an error, thusly:
```
Deploying devnet-jack-daniels network...
fatal: ref HEAD is not a symbolic ref
```

## How Has This Been Tested?

Locally

## Checklist:

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -53,7 +53,7 @@ done
 echo "Deploying $NETWORK_NAME network...";
 
 # TODO: this fails without an error if not at the HEAD of a given branch
-BRANCH=`git symbolic-ref --short -q HEAD`
+BRANCH=`git symbolic-ref --short HEAD`
 if [[ $BRANCH == "devnet-"* ]]; then
     if [[ $BRANCH != $NETWORK_NAME ]] && [[ $FORCED_DEPLOYMENT != 1 ]]; then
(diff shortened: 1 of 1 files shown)


nmarley (approved):
LGTM, 👍

bin/deploy
```diff
@@ -52,8 +52,8 @@ done
 
 echo "Deploying $NETWORK_NAME network...";
 
-# TODO: this fails without an error if not at the HEAD of a given branch
-BRANCH=`git symbolic-ref --short -q HEAD`
+# TODO: this fails if not at the HEAD of a given branch
```
nmarley, line 55: This PR is fine as-is (approved), but what is the TODO left here? Do you want it to not fail if not on a named branch? That would be easy to fix as you just disable errors and re-enable after the command.

One method:

```
set +e
git command here
set -e
```

Maybe cleaner alternative:

```
git command here || true
```

You could restore the `--quiet` option at this point and avoid the ugly error message. Sorry I didn't recognize this at the time, wasn't sure what the problem or goal was.

strophy: It's fine, I think it is good to fail with an error at least so I know where to look if I ever end up in this situation again. I don't think we should just bypass setting the variable, it's needed for the subsequent check.

PR #401, opened 2022-11-22
Zabbix monitoring for devnets

ktechmidas:
Zabbix monitoring can now be enabled on devnets

## What was done?
Zabbix security group rule has been added in terraform to allow 10050
Zabbix agent installation done via an Ansible role

## How Has This Been Tested?
Tested on devnet-ouzo

## Breaking Changes
None, zabbix will only install if zabbix=1 is in the config


strophy:
LGTM! But needs newlines at the end of a couple of files, and the FQCN style isn't used in ansible. This should be picked up by checks but CI didn't trigger for some reason.

strophy:
Closed, this is stale

PR #405, opened 2022-12-06
(do not merge): fee workaround

strophy:
## Issue being fixed or feature implemented

Core is spending masternode outputs when creating protx

## What was done?

Use a static fee address as workaround

## How Has This Been Tested?

On testnet

## Breaking Changes

None

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

PR #400, opened 2022-11-22
fix: collaterals being spent when creating protx

strophy:
## Issue being fixed or feature implemented

Collaterals are consumed as fees when creating protx transactions using a named wallet in Core 18

## What was done?

Work around the issue by funding a separate fee address

## How Has This Been Tested?

Currently testing on `devnet-feefix`

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #408, opened 2022-12-16
chore: fix loadwallet and handle misconfigured masternodes

strophy:

Supersedes #405 and #406 
#390 introduced use of the `loadwallet` command, which fails to lock masternode outputs present in the wallet on load. Deploy tool depends on these outputs being locked, so the issue results in masternode collaterals being consumed as fees in other transactions. This issue was handled with static fee addresses in #400, which introduced a new limitation on how many transactions the deploy tool could create without needing to wait for a block. We need to revert code from these two PRs and load both wallets on startup to ensure outputs are locked.

## What was done?

Revert #400 
Partially revert #390 
Introduce handling for misconfigured masternodes (ENABLED but with wrong service address, and cannot be banned because all nodes have wrong service address)

## How Has This Been Tested?

On `testnet` and `devnet-bintang`

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/roles/dashd/tasks/main.yml
+++ b/ansible/roles/dashd/tasks/main.yml
@@ -82,11 +82,12 @@
     mode: "+x"
 
 - name: Load/create faucet wallet on startup
-  ansible.builtin.lineinfile:
+  ansible.builtin.blockinfile:
     path: '/etc/dash.conf'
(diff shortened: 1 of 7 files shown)


shumkov (approved):
👍

PR #410, opened 2022-12-17
fix: elasticsearch cluster status yellow due to missing replicas

strophy:

The Elasticsearch `/_cluster/health` endpoint was returning `yellow` status due to missing replicas, even on networks where only a single logging node is present and replicas are therefore not possible. This was causing smoke tests to fail and turn red. We don't like yellow or red; green is our favourite colour.

## What was done?

Set `number_of_replicas: 0` if logging cluster size is 1.

## How Has This Been Tested?

On `devnet-bintang`

## Breaking Changes

None

## Checklist:


shumkov (approved):
👍

PR #411, opened 2022-12-19
feat: add quorum smoke tests

strophy:

Smoke tests should also check the status of quorums.

## What was done?

Add tests to check for:
- formation of the proper quorum types
- presence of chainlocks at the chain tip
- recency of the first quorum of an appropriate type for the target network
- presence of IS locks after creating a transaction

## How Has This Been Tested?

Locally, against `devnet-bintang`

## Breaking Changes

None

## Checklist:

### DIFF
--- a/test/smoke/quorum.js
+++ b/test/smoke/quorum.js
@@ -40,11 +40,11 @@ describe('Quorums', () => {
     const quorumList = {};
     const blockchainInfo = {};
     const firstQuorumInfo = {};
-    const transactionInfo = {};
-    let txid = '';
+    const memPoolEntry = {};
+    let instantsendTestTxid = '';
 
-    before('Collect chain lock and quorum list', function func() {
-      this.timeout(60000); // set mocha timeout
+    before('Collect chain lock and quorum list', () => {
+      // this.timeout(60000); // set mocha timeout
 
       const promises = [];
       for (const hostName of allHosts) {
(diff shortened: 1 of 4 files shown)


test/smoke/quorum.js
```diff
@@ -0,0 +1,181 @@
...
+
+const { inventory, network, variables } = getNetworkConfig();
+
+const allHosts = inventory.masternodes.hosts.concat(
+  inventory.wallet_nodes.hosts,
+  inventory.miners.hosts,
+  inventory.seed_nodes.hosts,
+);
+
+const quorumCheckTypes = {
```
shumkov, line 12: I guess the list should depend on the network, isn't? Or testnet and devnets have the same set?

strophy: Yeah, I'm not really sure how to configure this properly. dkgInterval is also the wrong measurement, quorums seem to get older than 24 blocks. How does this work?

shumkov: You probably need to have a map of which quorum types and the number of quorums should exist for testnet and devnet. Then use https://github.com/dashpay/dashcore-lib/blob/master/lib/deterministicmnlist/QuorumEntry.js#L335 to get params for specific quorum type.

strophy:
This will fail until `getMemPoolEntry` is available in `@dashpay/dashd-rpc`. Or I could commit add a patch in here?

test/smoke/quorum.js
strophy: How should I use this? It looks like it returns useful vars about the number of quorum members, should we add a check for that? Is there a similar function I can use to get the quorum lifetimes, or is it fine to use `dkgInterval * 1.5`?

shumkov: >  It looks like it returns useful vars about the number of quorum members, should we add a check for that?
I don't think so. If we have quorum in list it's already has min number of valid members.
>  Is there a similar function I can use to get the quorum lifetimes
I don't think so

shumkov (approved):
👍

PR #412, opened 2022-12-20
feat: reorder masternodes order

pshenmic:

Change the order of masternodes deploymen: masternode deployment should come after the registration. 
## What was done?

* Put masternodes deployment after the proreg tx registration
* Merged core and platform tasks into single task

## How Has This Been Tested?

Deployed network via this tool and ensured that all nodes became READY. Also checked elasticsearch logs after all.

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/deploy.yml
+++ b/ansible/deploy.yml
@@ -108,19 +108,7 @@
 
 # Start seeds
 
-- name: Set up core on seed nodes
-  hosts: seed_nodes
-  become: true
-  roles:
-    - role: dashd
-      dashd_listen: true
-      dashd_zmq: true
-      dashd_indexes: true
-      tags:
-        - dashd
-    - elastic_beats
-
-- name: Set up platform on seed nodes
+- name: Set up core and platform on seed nodes
   hosts: seed_nodes
   become: true
   pre_tasks:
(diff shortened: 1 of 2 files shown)


ansible/deploy.yml
```diff
@@ -190,47 +181,36 @@
...
       ansible.builtin.set_fact:
+        masternode: "{{ masternodes[inventory_hostname] }}"
         node: "{{ masternodes[inventory_hostname] }}"
       tags: always
       when: inventory_hostname in masternodes
     - name: Fail if no masternodes present
       ansible.builtin.fail:
         msg: Masternode not defined in network config
-      when: node is not defined
+      when: masternode is not defined
```
pshenmic, line 199: Where masternode OR node?

strophy: There's no real way to reach this logic branch, all masternodes and seed nodes are always defined on all networks. I find the names for these checks a bit weird as well, if we ever need to make a network without seed nodes or masternodes we will just comment out the tasks deploying roles to the relevant nodes instead. It's a lot of error boilerplate for no real benefit; I'll do a commit to show what I think can be removed, then I think we can merge this :+1:

PR #432, opened 2023-02-13
feat: terraform implementation for split arch networks

ktechmidas:
This implementation provides for split nodes between AMD and ARM

## What was done?
Terraform vars are now split into two (masternode{amd_count/arm_count}) and then combined later on into a local var for masternodes_count which includes both. Note the changes to services_output uses concat/join to ensure the masternodes list comes out correctly for ansible (which then won't care if we're AMD/ARM)

## How Has This Been Tested?
Tested by deploying devnets

## Breaking Changes
masternode_count in the tfvars is now split into
masternode_amd_count
masternode_arm_count

Please ensure this is configured on existing devnets (and testnet) before deployment


terraform/aws/main.tf
```diff
@@ -201,8 +218,8 @@ locals {
 }
 
 resource "random_shuffle" "dns_ips" {
-  input        = concat(aws_instance.masternode.*.public_ip)
-  result_count = length(concat(aws_instance.masternode.*.public_ip)) > local.dns_record_length ? local.dns_record_length : length(concat(aws_instance.masternode.*.public_ip))
+  input        = concat(aws_instance.masternode_amd.*.public_ip)
+  result_count = length(concat(aws_instance.masternode_amd.*.public_ip)) > local.dns_record_length ? local.dns_record_length : length(concat(aws_instance.masternode_amd.*.public_ip))
```
strophy, line 222: Does this mean only IPs from AMD hosts can appear in the DNS list?

ktechmidas: Fixed

strophy (approved):
Great work mate! :rocket:

PR #434, opened 2023-02-21
chore: refactor generate blocks

strophy:

`mn_init` was already unreliable when running against a single list of masternodes, and in >50% of cases needed to be run twice (manually) to make sure enough blocks were actually mined to confirm transactions. This behavior was exacerbated when running `mn_init` twice in a single play against lists registering `mn` and `hpmn`, and it became necessary to refactor the entire logic of how we produce blocks on different networks.

## What was done?

- Simplify `generate_blocks` role to remove any distinction between devnets and testnet, using conditional logic around block mining instead
- Refactor `generate_blocks` task to mine blocks with an explicitly stated goal to `generate` one of:
  - a target `balance`
  - a target number of `blocks`
  - one-block `confirmations` for a given list of txids (this is the most common case in which `mn_init` would unpredictably fail, caused by miners omitting some transactions in mempool for reasons)

## How Has This Been Tested?

On `devnet-xxxx` 

## Breaking Changes

None

## Checklist:


ansible/roles/generate_blocks/tasks/main.yml
```diff
@@ -1,20 +1,71 @@
...
-  ansible.builtin.include_tasks: wait_for_blocks.yml
-  when: dash_network not in generate_networks
+# Option: blocks
+
+- name: Wait for {{ num_blocks ~ ' block(s)' }}
+  ansible.builtin.command: dash-cli getblockcount
+  when: generate == 'blocks'
+  register: getblockcount_result
+  until: getblockcount_result.stdout | int >= start_block_count.stdout | int + num_blocks | int
+  retries: 200
```
shumkov, line 39: Shouldn't this value depend on how many blocks / dash we generate?

ansible/roles/mn_fund_collateral/tasks/main.yml
```diff
@@ -35,9 +35,9 @@
   ansible.builtin.include_role:
     name: generate_blocks
   vars:
-    num_blocks: 0
-    balance_needed: "{{ (masternode_names | length - collateral_ok_count | int) * (funding_amount | int / 100000) | int }}"
-  when: not collateral_all_ok and (faucet_balance.stdout | int < balance_needed | int)
+    generate: balance
+    target_balance: "{{ (masternode_names | length - collateral_ok_count | int) * (funding_amount | int / 100000) | int }}"
+  when: not collateral_all_ok and (faucet_balance.stdout | int < target_balance | int)
```
shumkov, line 40: Don't you have such check already inside?

ansible/roles/generate_blocks/tasks/main.yml
strophy: Good point, but I think we would need to know the current block subsidy in order to calculate this? 

I've never encountered an insufficient balance condition because we mine 4032 blocks with minimum difficulty on all networks anyway, so I don't think it really matters.

ansible/roles/generate_blocks/tasks/main.yml
```diff
@@ -1,20 +1,71 @@
...
+
+# Stop mining
+
+- name: Stop generating blocks
+  ansible.builtin.systemd:
+    name: 'dashd-generate-miner'
+    state: stopped
+    enabled: false
+    masked: false
+  when: generate == 'confirmations' and dash_network in generate_networks
```
shumkov, line 71: Do we stop it for confirmations only?

ansible/roles/generate_blocks/tasks/main.yml
shumkov: Ok, so if you think it will never happen in any possible use cases you can keep like this I guess

ansible/roles/mn_fund_collateral/tasks/main.yml
strophy: Where? You mean [this check](https://github.com/dashpay/dash-network-deploy/pull/434/files#diff-c83d51b081e08f1a45847f902e089cf58b36a0c879981e56e1bbd339d5e6bd52R48)?

ansible/roles/generate_blocks/tasks/main.yml
strophy: Good catch! This would have left the wallet node mining if one of the other conditions was used :see_no_evil:

ansible/roles/mn_fund_collateral/tasks/main.yml
shumkov: Yep

strophy: One check is whether it is necessary to enter the `generate_blocks` function, the other check is when to stop generating the blocks (after determining it was necessary to start in the first place).

There's no need to start and stop the systemd mining service just to run the balance check.

shumkov (approved):
👍

PR #438, opened 2023-02-23
fix: move import wallet and check

ktechmidas:
Fixes the import wallet issue on redeployment by moving the import above starting (so it is already there when we start)

If there is no wallet, it will *not* add to the configs. However it will then create (and import) the wallets on first loadup. 

There is a small issue where if we restart the wallet node *after* first deployment, but have not done subsequent deployments, the wallet won't be there. This is minor, and can be fixed by simply loading the wallet, I do not expect this to come up very often (mostly for failed deployments where we do a manual restart)

## How Has This Been Tested?
devnet-infratesting

### DIFF
--- a/ansible/roles/dashd/tasks/main.yml
+++ b/ansible/roles/dashd/tasks/main.yml
@@ -81,6 +81,20 @@
     path: /etc/cron.daily/logrotate
     mode: "+x"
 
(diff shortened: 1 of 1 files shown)


strophy (approved):
:+1:

PR #458, opened 2023-03-28
feat: EIP initial implementation

ktechmidas:
Change public IP's to elastic IP addresses for HPMNs. This ensures HPMNs will keep their IP address through rebooting/changing things/etc

## What was done?
New blocks for EIPs for both ARM/AMD, turn public_ip_allocation off for those nodes and add an EIP instead.

## How Has This Been Tested?
devnet-eip

## Breaking Changes
~None, AFAIK~
Requires Terraform version 1.4.4 due to bug in previous versions.

## Checklist:


terraform/aws/main.tf
```diff
@@ -240,22 +240,22 @@ resource "random_shuffle" "dns_ips" {
   input        = concat(
     aws_instance.masternode_amd.*.public_ip,
     aws_instance.masternode_arm.*.public_ip,
-    aws_instance.hp_masternode_amd.*.public_ip,
-    aws_instance.hp_masternode_arm.*.public_ip
+      aws_eip.hpmn_arm_eip.*.public_ip,
+      aws_eip.hpmn_amd_eip.*.public_ip
```
strophy, line 244: It's definitely necessary to reference the EIP itself here right? It's not possible to reference the instance instead, and let the graph figure out that the association is necessary before it gets to this step? After all the EIP *is* the `public_ip` associated with the instance.

ktechmidas: Definitely necessary unfortunately, the public_ip within the instance itself is the one AWS gives it before the EIP, not the EIP.

terraform/aws/instances.tf
```diff
@@ -317,6 +336,29 @@ resource "aws_instance" "hp_masternode_arm" {
...
+}
+
+
+output "hp_masternode_arm_eip" {
+  value = aws_eip.hpmn_arm_eip.*.public_ip
+}
+
+output "hp_masternode_amd_eip" {
+  value = aws_eip.hpmn_amd_eip.*.public_ip
+}
```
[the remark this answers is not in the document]
ktechmidas: Was for debugging, removed now.

strophy (approved):
Works for me :rocket:

PR #472, opened 2023-04-13
feat!: Make EIP and EIP association optional based on a variable

ktechmidas:
<pre>
feat!: Make EIP and EIP association optional based on a variable

This change allows the option to create or not create Elastic IPs and EIP associations based on a variable. This provides more flexibility when deploying resources in different scenarios where EIPs are not required or when public IPs should be assigned directly.

## What was done?
- Added a new `create_eip` variable to control the creation of EIPs and EIP associations
- Updated `aws_instance` resource to conditionally assign a public IP based on the `create_eip` variable
- Updated `aws_eip_association` resource to conditionally create the association based on the `create_eip` variable

## How Has This Been Tested?
Tested by deploying resources with the `create_eip` variable set to `true` and `false` and verifying the expected behavior.

## Breaking Changes
This change introduces a new variable `create_eip`, which may require updating existing configurations to include this variable. However, the variable is set to `true` by default, so it should not affect current deployments unless explicitly set to `false`.

## Checklist:

</pre>

### DIFF
--- a/terraform/aws/instances.tf
+++ b/terraform/aws/instances.tf
@@ -240,7 +240,7 @@ resource "aws_instance" "masternode_arm" {
 
 resource "aws_eip" "hpmn_arm_eip" {
   instance = null
-  count = var.hp_masternode_arm_count
+  count = var.create_eip ? var.hp_masternode_arm_count : 0
   
   tags = {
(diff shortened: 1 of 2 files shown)


shumkov (approved):
👍

PR #476, opened 2023-04-18
fix: is lock smoke test expected 'false' to equal 'true'

strophy:

IS tests would time out on testnet with:
```
      AssertionError: expected 'false' to equal 'true'
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/smoke/quorum.js:261:76)
```
on around half the nodes.

## What was done?

Increase wait time before checking for instantsend lock on the transaction

## Checklist:

### DIFF
--- a/test/smoke/quorum.js
+++ b/test/smoke/quorum.js
@@ -191,10 +191,10 @@ describe('Quorums', () => {
     });
 
     before('Collect instantsend info', async () => {
-      // Wait two seconds here before checking for IS locks
+      // Wait five seconds here before checking for IS locks
       // TODO: implement this.slow() and await IS ZMQ message to mark test response speed yellow/red
       await new Promise((resolve) => {
-        setTimeout(resolve, 4000);
+        setTimeout(resolve, 5000);
       });
 
       const promises = [];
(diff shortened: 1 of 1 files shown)


test/smoke/quorum.js
```diff
@@ -194,7 +194,7 @@ describe('Quorums', () => {
       // Wait two seconds here before checking for IS locks
       // TODO: implement this.slow() and await IS ZMQ message to mark test response speed yellow/red
       await new Promise((resolve) => {
-        setTimeout(resolve, 2000);
+        setTimeout(resolve, 4000);
```
PastaPastaPasta, line 197: 4 seconds may not be enough. ISLocks can sometimes take longer. Probably up to 10 seconds is reasonable. Additionally your comment is now incorrect

strophy: Good point. It's a trade-off with waiting because if a block is mined in those 10s, then the test will fail as well. I think implementing ZMQ here is probably a better way to design the test.

ktechmidas (approved):
LGTM

strophy:
Going with 5s here because it currently easily passes with this threshold, so if this test starts failing then the current (apparently normal) IS lock time of <5s serves as a good baseline for devnets/testnet at least.

PR #481, opened 2023-04-24
fix: eip nodes being recreated on each tf run

strophy:

Nodes with EIPs were being destroyed and recreated each time terraform ran because `associate_public_ip_address` apparently changes from `false` (the value we set) to `true` after the EIP is associated. 

## What was done?

Add `associate_public_ip_address` to the `ignore_changes` lifecycle

## How Has This Been Tested?

On `testnet` and `devnet-two-islands`, no changes against both

## Breaking Changes

None

## Checklist:


ktechmidas:
Please don't merge this, it's incorrect and there is something else going on here:

setting associate_public_ip_address on creation creates (or doesn't) a public IP address for the instance, this is entirely different from an EIP: https://www.quora.com/Whats-the-difference-between-a-public-IP-and-an-elastic-IP-in-AWS-EC2

This worked fine for a month or so before this PR: https://github.com/dashpay/dash-network-deploy/pull/472

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-network-iface-embedded.html#aws-properties-ec2-network-iface-embedded-associatepubip

I suspect instead that we're being overridden by subnet settings with map_public_ip_on_launch or something similar. An instance can have two IP addresses, one public that changes every time which is then overridden by an EIP, this is how it should work and has been working

ktechmidas (changes requested):
Please see comment

ktechmidas:
Fixed here: https://github.com/dashpay/dash-network-deploy/pull/482

Please do not merge this one

PR #482, opened 2023-04-24
fix: always set associated IP to be true

ktechmidas:
Issue described in PR https://github.com/dashpay/dash-network-deploy/pull/481 - this is a better fix than ignoring lifecycle changes and supersedes that PR

## What was done?
Always set associate public IP to true even if we are going to later replace with an EIP. Terraform does not track state changes correctly here (breaking standards with AWS CloudFormation which does but I digress...)

Terraform has a few long standing issues about this: 

https://github.com/hashicorp/terraform-provider-aws/issues/343

https://github.com/hashicorp/terraform/issues/9811

## How Has This Been Tested?
Devnet

## Breaking Changes
Possibly previous networks with HPMNs may need to redeploy.

## Checklist:

- [ x] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have added or updated relevant unit/integration/functional/e2e tests
- [x ] I have made corresponding changes to the documentation

- [x ] I have assigned this pull request to a milestone

### DIFF
--- a/terraform/aws/instances.tf
+++ b/terraform/aws/instances.tf
@@ -264,7 +264,7 @@ resource "aws_instance" "hp_masternode_amd" {
   instance_type        = "t3.medium"
   key_name             = aws_key_pair.auth.id
   iam_instance_profile = aws_iam_instance_profile.monitoring.name
-  associate_public_ip_address = !var.create_eip
+  associate_public_ip_address = true
 
   vpc_security_group_ids = [
     aws_security_group.default.id,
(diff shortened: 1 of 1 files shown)


strophy (approved):
Works for me :+1: But will this result in AWS associating non-EIP public addresses with the instances in addition to the EIPs?

ktechmidas:
> Works for me 👍 But will this result in AWS associating non-EIP public addresses with the instances in addition to the EIPs?

Yes and no, once the EIP gets added the original address goes away. It doesn't really matter in the grand scheme of things

PR #495, opened 2023-05-09
fix!: platfrom block proposals too slow

shumkov:

Drive is too slow to create proposals in time so validators are voting nil due to timeout

## What was done?

- Set genesis tenderdash timeouts

## How Has This Been Tested?

On testnet

## Breaking Changes

Previous platform chaines won't be valid since genesis config is changed.

## Checklist:

### DIFF
--- a/ansible/roles/dashmate/templates/dashmate.json.j2
+++ b/ansible/roles/dashmate/templates/dashmate.json.j2
@@ -171,11 +171,11 @@
               "initial_core_chain_locked_height": {{ platform_initial_core_chain_locked_height | int }},
               "consensus_params": {
                 "timeout": {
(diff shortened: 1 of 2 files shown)


strophy:
Are these the timeouts within which a node must be able to complete a given action? Can this be used for benchmarking or to find the minimum required spec for a HPMN?

shumkov:
> Are these the timeouts within which a node must be able to complete a given action? 
Yep, kinda.

For benchmarking we need to implement instrumentation for Drive and Tenderdash and use some testing framework so we can test different scenarios and load. The instrumentation should help us to analyze performance and required hardware requirements.

strophy (approved):

PR #503, opened 2023-05-16
refactor: start dashmate services depending on status

pshenmic:

Start or restart dashmate services depending on the dashmate status and dashmate update command.

https://github.com/dashpay/dash-network-deploy/issues/513

## What was done?

* Added smart start / restart of dashmate services depending on service statuses, image updates, package version or config changes
* Fixed config idempotency (on template render)

## How Has This Been Tested?

On devnet

## Breaking Changes

None

## Checklist:

### DIFF
--- a/ansible/roles/dashmate/tasks/install.yml
+++ b/ansible/roles/dashmate/tasks/install.yml
@@ -17,3 +17,4 @@
   ansible.builtin.apt:
     deb: "{{ release_info.json.assets | selectattr('name', 'search', '_' + deb_architecture.stdout + '.deb') | map(attribute='browser_download_url') | first }}"
     allow_downgrade: true
+  register: dashmate_install_result
--- a/ansible/roles/dashmate/tasks/main.yml
+++ b/ansible/roles/dashmate/tasks/main.yml
@@ -141,8 +141,19 @@
     chdir: '{{ dashmate_cwd }}'
   register: dashmate_obtain
   changed_when: dashmate_obtain.rc == 0
-  when: dashmate_platform_enable and dashmate_platform_dapi_envoy_ssl_provider ==
-    'zerossl' and dashmate_platform_dapi_envoy_ssl_provider_config_zerossl_id is not defined
+  when: >
+    dashmate_platform_enable and
+    dashmate_platform_dapi_envoy_ssl_provider == 'zerossl' and
+    dashmate_platform_dapi_envoy_ssl_provider_config_zerossl_id is not defined
+
+- name: Change dashmate image to avoid 'pull access denied for dashmate_local'
+  ansible.builtin.command: "{{ dashmate_cmd }} config set dashmate.helper.docker.image dashpay/dashmate-helper:latest-dev"
+  become: true
+  become_user: dashmate
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  changed_when: true
+  when: dashmate_branch is defined
 
 - name: Update dashmate images
   ansible.builtin.command: "{{ dashmate_cmd }} update --verbose"
--- a/ansible/roles/dashmate/tasks/main.yml
+++ b/ansible/roles/dashmate/tasks/main.yml
@@ -190,7 +190,7 @@ Core Status: {{ core_docker_status }}\n
 Tenderdash Status: {{ tenderdash_docker_status }}\n
 Core Images Updated: {{ core_images_updated }}\n
 Platform Images Updated: {{ platform_images_updated }}\n
-Dashmate Package Updated: {{ dashmate_package_changed }}\\n
+Dashmate Package Updated: {{ dashmate_package_changed }}\n
 Dashmate Config Changed: {{ dashmate_config_changed }}
 "
 
--- a/ansible/roles/dashmate/tasks/main.yml
+++ b/ansible/roles/dashmate/tasks/main.yml
@@ -94,10 +94,6 @@
   ansible.builtin.import_tasks: ./install.yml
   when: dashmate_version is defined and dashmate_branch is not defined
 
-- name: print
-  ansible.builtin.debug:
-    var: dashmate_version
-
 - name: Write dashmate config file
   vars:
     template_bootstrap_peers: "{{ groups.seed_nodes | reject('equalto', inventory_hostname) }}"
@@ -108,20 +104,15 @@
     group: '{{ dashmate_group }}'
     mode: "0644"
   register: template_result
-  diff: true
-#
-#- name: print
-#  ansible.builtin.debug:
-#    var: template_result
-#
-#- name: Generate self-signed SSL certificate for DAPI
-#  ansible.builtin.import_tasks: ./ssl/self_signed.yml
-#  when: dashmate_platform_enable and dashmate_platform_dapi_envoy_ssl_provider == 'self-signed'
-#
-#- name: Obtain ZeroSSL certificate for DAPI
-#  ansible.builtin.import_tasks: ./ssl/zerossl.yml
-#  when: dashmate_platform_enable and dashmate_platform_dapi_envoy_ssl_provider == 'zerossl'
-#
+
+- name: Generate self-signed SSL certificate for DAPI
+  ansible.builtin.import_tasks: ./ssl/self_signed.yml
+  when: dashmate_platform_enable and dashmate_platform_dapi_envoy_ssl_provider == 'self-signed'
+
+- name: Obtain ZeroSSL certificate for DAPI
+  ansible.builtin.import_tasks: ./ssl/zerossl.yml
+  when: dashmate_platform_enable and dashmate_platform_dapi_envoy_ssl_provider == 'zerossl'
+
 - name: Export current dashmate config
   ansible.builtin.shell: "{{ dashmate_cmd }} config envs > {{ dashmate_home }}/.env"
   become: true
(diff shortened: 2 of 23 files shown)


ansible/roles/dashmate/tasks/main.yml
```diff
@@ -173,8 +173,20 @@
...
-#   to get the status from the response
+- name: Get status
+  ansible.builtin.command: "{{ dashmate_cmd }} status --format=json"
+  become: true
+  become_user: dashmate
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_status
+  tags:
+    - skip_ansible_lint
```
strophy, line 184: Why do we skip lint on this task?

strophy (changes requested):
What are the possible values for `service_status` other than `up` and `stopped`? Will the value still be `stopped` if we are on the initial run, the `dashmate start` command has never run yet and the containers don't exist? Should we add a fallback in case the status command crashes, or should deploy tool just exit in that case?

ansible/roles/dashmate/tasks/main.yml
pshenmic: Its complaing about no-changed-when and that command should not change anything, but its not changing anything, it just registers the variable

strophy (approved):

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -173,8 +173,24 @@
...
+- name: Get status
+  ansible.builtin.command: "{{ dashmate_cmd }} status --format=json"
+  become: true
+  become_user: dashmate
+  changed_when: true
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_status
+
+- name: Get service status
```
shumkov, line 185: it might be situation that core is started but platform is not (after platform destroy) so you need to check both, drive and core. If core is started but not a platform you should use `start --platform`

shumkov (changes requested):

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -173,8 +173,53 @@
...
+    tenderdash_service_status: "{{ dashmate_status.stdout | from_json | json_query('platform.tenderdash.serviceStatus') }}"
+
+- name: Fail if dashmate node in inconsistent state
+  ansible.builtin.fail:
+    msg: Dashmate status returned invalid service status
+  when: >
+    (core_service_status != 'up' and core_service_status != 'stopped') or
+    (tenderdash_service_status != 'up' and tenderdash_service_status != 'stopped')
+
+# start platform when (core is up and platform is down)
```
shumkov, line 197: Don't we need to restart Core as well? What if dashmate code is updated or config is changed?

We can do restart logic smarter:
1. Images are changes: restart core if core image changed and restart platform if platform images are changed
2. dashmate config is changed
3. dashmate is changed

shumkov (changes requested):

ansible/roles/dashmate/tasks/main.yml
strophy: Right, yesterday I had difficulty updating from Platform 0.24.2 to 0.24.4 because `docker compose restart` does not pull and use newer images, `docker compose up -d` is necessary for this. We definitely need a way of ensuring the necessary containers are restarted (or stopped, updated and started) when config changes. But changing the dashmate version probably shouldn't cause core to restart - this usually changes at the same time as the platform version only.

shumkov: The full logic must be:
* check services status
  * core is running
  * drive is running
* start ALL when (core is down and platform is down)
* start platform when (core is up and platform is down)
* determine what's need to be restarted:
  * any of platform images are updated: platform
  * core / sentinel images are updated: all
  * dashmate is changed: all
  * dashmate config is changed: all
* restart platform (platform is not down and platform needs to be restarted)
* restart all (all needs to be restarted)

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -173,8 +173,53 @@
...
+  become_user: dashmate
+  changed_when: true
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_status
+
+- name: Get service status
+  ansible.builtin.set_fact:
+    core_service_status: "{{ dashmate_status.stdout | from_json | json_query('core.serviceStatus') }}"
+    tenderdash_service_status: "{{ dashmate_status.stdout | from_json | json_query('platform.tenderdash.serviceStatus') }}"
```
shumkov, line 188: It seems we need to get all platform service statuses otherwise it will be inconsistency and conflict with the check in dashmate https://github.com/dashpay/platform/blob/2e984df0302fa94bdf187bd893328ec58f390aa0/packages/dashmate/src/docker/DockerCompose.js#L75

We using this function in start and restart commands I believe

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -148,30 +149,113 @@
...
+  become_user: dashmate
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_start_platform
+  when:
+    - core_docker_status == 'running'
+    - tenderdash_docker_status == 'not_started'
+  changed_when: dashmate_start_platform.rc == 0
+
+# Restart logic
```
strophy, line 227: Currently there is no way to restart only the core container, right?

strophy:
This looks really good, lots of useful changes here. Does it currently depend on a minimum version of dashmate for the restart platform commands to work properly? Or do we need to do a release still?

pshenmic:
No, required things are already included in the last release (v0.24.14).

We still could have some idempotency defects in rare cases, but they will be resolved once this https://github.com/dashpay/platform/issues/1181 is done.

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -148,30 +149,113 @@
...
   args:
     chdir: '{{ dashmate_cwd }}'
-  register: dashmate_restart
-  changed_when: dashmate_restart.rc == 0
-  ignore_errors: true
+  register: dashmate_status
+
+- name: Get service status
+  ansible.builtin.set_fact:
+    core_docker_status: "{{ dashmate_status.stdout | from_json | json_query('core.dockerStatus') }}"
```
shumkov, line 163: What if tenderdash is stopped it Drive or DAPI is started?

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -148,30 +149,113 @@
...
+    - dashmate_package_changed is false and dashmate_config_changed is false
+  changed_when: dashmate_restart.rc == 0
+
+- name: Restart dashmate services
+  ansible.builtin.command: "{{ dashmate_cmd }} restart --verbose"
+  become: true
+  become_user: dashmate
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_restart
```
shumkov, line 252: Better use different names

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -148,30 +149,113 @@
...
+
+- name: Restart dashmate platform services
+  ansible.builtin.command: "{{ dashmate_cmd }} restart --platform --verbose"
+  become: true
+  become_user: dashmate
+  args:
+    chdir: '{{ dashmate_cwd }}'
+  register: dashmate_restart
+  when:
+    - core_docker_status == 'running' and tenderdash_docker_status == 'running'
```
shumkov, line 241: Are they working as AND or OR?

ansible/roles/dashmate/templates/dashmate.json.j2
```diff
@@ -1,5 +1,5 @@
 {
-  "configFormatVersion": "{{ dashmate_version | default('999.0.0') }}",
+  "configFormatVersion": "{{ dashmate_version | default(dashmate_config_version | default('999.0.0')) }}",
```
shumkov, line 2: Why do you we want to configure this?

shumkov (changes requested):

ansible/roles/dashmate/tasks/install.yml
```diff
@@ -17,3 +17,4 @@
   ansible.builtin.apt:
     deb: "{{ release_info.json.assets | selectattr('name', 'search', '_' + deb_architecture.stdout + '.deb') | map(attribute='browser_download_url') | first }}"
     allow_downgrade: true
+  register: dashmate_install_result
```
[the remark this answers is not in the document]
pshenmic: Yes, in the `dashmate_package_changed`

ansible/roles/dashmate/templates/dashmate.json.j2
pshenmic: Because when we install from branch, we don't have dashmate_version available and it will back off to `999.0.0` which will break template idempotency. On the first dashmate run, config version will be overwritten with the package default version.

ansible/roles/dashmate/tasks/main.yml
pshenmic: As AND. We only want to restart if all services are currently running

ansible/roles/dashmate/tasks/main.yml
pshenmic: What do you mean? Should I rather check by drive container?

ansible/roles/dashmate/templates/dashmate.json.j2
shumkov: I see. Yes you right. Would it be better to get dashmate version form dashmate itself (`dashmate --version` or `package.json`) so you won't need to configure it manually. `999.0.0` workaround won't be necessary as well.

ansible/roles/dashmate/tasks/main.yml
shumkov: If one of them started?

ansible/roles/dashmate/tasks/main.yml
pshenmic: Is there a case for it?

ansible/roles/dashmate/tasks/main.yml
pshenmic: I'm worried that it will add more fragility to the code. We actually don't have them stopped separately from each other. Tenderdash container will be either in `not_started` state or `running`. If there are any issues happening, it will remain in `restarting`, `exited`, or any other states, and tasks will be skipped

shumkov (changes requested):

pshenmic:
`What will happen if tenderdash is stopped but any of platform services is running? DAPI for example?`

It will try to start platform services

ansible/roles/dashmate/tasks/main.yml
strophy: e.g. an update to core is released, and we should update that container without changing anything on platform

pshenmic: No, this is not included in the task and we need to implement it in dashmate first. In this case everything will be restarted for now

strophy (approved):

PR #504, opened 2023-05-16
feat: ELB & certs for seed addresses

ktechmidas:
The feature here is the creation and configuration of an AWS Elastic Load Balancer (ELB), ACM Certificate, Route53 Record and Security Group using Terraform. This setup is required to make seed-*.testnet.networks.dash.org accessible over HTTPS since the old method will not work while DAPI has differing certificates.

## What was done?
New ELB for seeds with the correct certificates, adding HPMNs as backends on port 3001. 

## How Has This Been Tested?
devnet-dude

## Breaking Changes
N/A

## Checklist:

### DIFF
--- a/terraform/aws/main.tf
+++ b/terraform/aws/main.tf
@@ -149,14 +149,14 @@ resource "aws_elb" "web" {
 }
 
 resource "aws_lb" "seed" {
-  name                       = "${var.public_network_name}-alb-seed"
+  name                       = "${var.public_network_name}-nlb-seed"
   internal                   = false
   load_balancer_type         = "network"
   subnets                    = aws_subnet.public.*.id
   enable_deletion_protection = false
 }
 
-resource "aws_lb_listener" "seed_listener" {
+resource "aws_lb_listener" "seed" {
   load_balancer_arn = aws_lb.seed.arn
   port              = var.dapi_port
   protocol          = "TLS"
--- a/terraform/aws/main.tf
+++ b/terraform/aws/main.tf
@@ -112,10 +112,10 @@ resource "aws_elb" "web" {
   }
 
   listener {
-    instance_port     = var.faucet_port
-    instance_protocol = "http"
-    lb_port           = var.faucet_https_port
-    lb_protocol       = "https"
+    instance_port      = var.faucet_port
+    instance_protocol  = "http"
+    lb_port            = var.faucet_https_port
+    lb_protocol        = "https"
     ssl_certificate_id = aws_acm_certificate_validation.faucet.certificate_arn
   }
 
@@ -149,11 +149,11 @@ resource "aws_elb" "web" {
 }
 
 resource "aws_lb" "seed" {
-  name               = "${var.public_network_name}-alb-seed"
-  internal           = false
-  load_balancer_type = "application"
-  security_groups    = [aws_security_group.seed.id]
-  subnets            = aws_subnet.public.*.id
+  name                       = "${var.public_network_name}-alb-seed"
+  internal                   = false
+  load_balancer_type         = "application"
+  security_groups            = [aws_security_group.seed.id]
+  subnets                    = aws_subnet.public.*.id
   enable_deletion_protection = false
 }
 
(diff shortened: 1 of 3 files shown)


terraform/aws/main.tf
```diff
@@ -148,6 +148,74 @@ resource "aws_elb" "web" {
...
+  internal           = false
+  load_balancer_type = "application"
+  security_groups    = [aws_security_group.sg-seed.id]
+  subnets            = aws_subnet.public.*.id
+  enable_deletion_protection = false
+}
+
+resource "aws_lb_listener" "listener" {
+  load_balancer_arn = aws_lb.seed-elb.arn
+  port              = 443
```
shumkov, line 162: It should be dapi_port variable

terraform/aws/main.tf
```diff
@@ -148,6 +148,74 @@ resource "aws_elb" "web" {
...
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.tg-seed.arn
+  }
+}
+
+resource "aws_lb_target_group" "tg-seed" {
+  name     = "${var.public_network_name}-tg-seed"
+  port     = 3001
```
shumkov, line 175: It should be dapi_port as well

terraform/aws/security_groups.tf
```diff
@@ -444,3 +444,28 @@ resource "aws_security_group" "vpn" {
...
 }
+
+resource "aws_security_group" "sg-seed" {
+  name        = "seed-1"
+  description = "Allow inbound traffic on port 443 and all outbound traffic"
+  vpc_id      = aws_vpc.default.id
+
+  ingress {
+    description = "TLS from VPC"
+    from_port   = 443
```
shumkov, line 455: dapi_port?

shumkov (dismissed):

terraform/aws/main.tf
```diff
@@ -148,6 +148,74 @@ resource "aws_elb" "web" {
   }
 }
 
+resource "aws_lb" "seed-elb" {
```
strophy, line 151: Snake case for variable names instead of kebab case

terraform/aws/security_groups.tf
```diff
@@ -444,3 +444,28 @@ resource "aws_security_group" "vpn" {
     DashNetwork = terraform.workspace
   }
 }
+
+resource "aws_security_group" "sg-seed" {
```
strophy, line 448: Other security groups are not named with `sg-` prefix, follow naming convention

terraform/aws/main.tf
```diff
@@ -256,14 +324,47 @@ resource "random_shuffle" "dns_ips" {
...
 
 # `seed-n` type addresses are dapi endpoints
-resource "aws_route53_record" "masternodes" {
-  zone_id = data.aws_route53_zone.main_domain[0].zone_id
-  name    = "seed-${count.index + 1}.${var.public_network_name}.${var.main_domain}"
-  type    = "A"
-  ttl     = "300"
-  records = random_shuffle.dns_ips.result
 
-  count = length(var.main_domain) > 1 ? 5 : 0
```
strophy, line 266: Can we keep this method of iteration? The new code is less flexible in the number of R53 records created.

strophy (changes requested):
Nice work, general design looks good. Left a few code comments. Would a network load balancer be sufficient here, why do we need the ALB?

ktechmidas:
@strophy could have gone either way, but if DAPI could, for example, tell us if a node is banned it could be removed, A/B testing could be done etc

An ALB is only a tiny bit more expensive and gives us more options

plus:

> Both Classic Load Balancers and Application Load Balancers use connection multiplexing, but Network Load Balancers do not. Therefore, your targets can receive more TCP connections behind a Network Load Balancer. Be sure that your targets are prepared to handle the volume of connection requests they might receive.

terraform/aws/security_groups.tf
```diff
@@ -444,3 +444,28 @@ resource "aws_security_group" "vpn" {
...
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
```
strophy, line 466: What is this for?

terraform/aws/security_groups.tf
```diff
@@ -444,3 +444,28 @@ resource "aws_security_group" "vpn" {
...
   }
 }
+
+resource "aws_security_group" "seed" {
+  name        = "seed-1"
+  description = "Allow inbound traffic on DAPI port and all outbound traffic"
+  vpc_id      = aws_vpc.default.id
+
+  ingress {
+    description = "TLS from VPC"
```
strophy, line 454: Isn't this from everywhere?

strophy (approved):
Nice work mate!

PR #520, opened 2023-06-25
feat: initial multi-stage deployment

ktechmidas:
This will make the deployment multi-staged, so we can run stage2 after chainlocks are formed. This is important because we need chainlocks in order to:

1) Anticipate this change: https://github.com/dashpay/dash/pull/5444 and do a two-staged spork enable
2) Start mixing
3) Start proposing as we need superblocks for this which aren't usually enabled until ~4200

We could check quorums for the above, but since quorums are required for chainlocks I thought it better to check for chainlocks then we don't need to guess if the quorums are good.

## What was done?
Changed /bin/deploy to have a new variable called STAGE where we can give it the stage we want to deploy.
Created continue.yml with a basic chainlock check

## How Has This Been Tested?
devnet-yankee

## Breaking Changes
None, leaving out "stage" will deploy stage1 as normal.


ansible/continue.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Execute getbestchainlock
+  hosts: wallet_nodes
+  tasks:
+    - name: Run dash-cli getbestchainlock
+      ansible.builtin.command: dash-cli getbestchainlock
+      register: chainlock_output
+      ignore_errors: true
+      no_log: true
```
strophy, line 9: Why `no_log`?

ansible/continue.yml
```diff
@@ -0,0 +1,14 @@
...
+
+    - name: Check for chainlock
+      ansible.builtin.fail:
```
strophy, line 12: `fail` shouldn't be necessary if we just remove the `ignore_errors` from previous step and/or add `failed_when`. But even that shouldn't be necessary, it will fail anyway when `chainlock_output.rc != 0` right?

ansible/continue.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Execute getbestchainlock
+  hosts: wallet_nodes
+  tasks:
+    - name: Run dash-cli getbestchainlock
+      ansible.builtin.command: dash-cli getbestchainlock
+      register: chainlock_output
+      ignore_errors: true
+      no_log: true
+
```
strophy, line 10:
```suggestion
changed_when: chainlock_output.rc == 0`

```

strophy (changes requested):
I'm not sure having `continue.yml` launch as a sub-stage of the deploy command makes sense, we either need to add a lot of error handling here (e.g. what if we deploy stage 2, but specify `-i` flag? I think it would be simpler to remove the `case` statement and just create a new top level executable script like `bin/start-stage-2-services`, except with a better name than that.

ktechmidas:
-i wouldn't affect stage2, it will just be ignored @strophy since it only actually does anything with -i in stage1

image

ansible/continue.yml
ktechmidas: In future, we will need to check exactly why chainlocks aren't there. IE: are the sporks on? If no... turn them on then fail, if yes, just fail.

This is just a template for that behavior from v20

strophy (approved):
Good start, let's see how this goes when we start adding functions

PR #528, opened 2023-07-17
Mixing code for networks

ktechmidas:
Mobile team requested mixing to continuously run on devnets

## What was done?
1) loadwallet for faucet and mno wallets now autoloads in future (so we don't need to reload everytime we restart wallet nodes)
2) Introduces new var (coinjoin_wallet_count) which specifies how many wallets should be created on wallet nodes
3) Makes the wallets in a coinjoin-wallet-n format
4) Adds dash to wallets from faucet wallet
5) Starts mixing

## How Has This Been Tested?
Tested on devnet-yankee

## Breaking Changes
None

## Checklist:


ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -116,3 +116,8 @@ port={{ dashd_port }}
 {% if dashd_seednode is not none and not dashd_isseednode %}
 addnode={{ dashd_seednode }}:{{ dashd_port }}
 {% endif %}
+
+{% if dashd_walletnode is not none and coinjoin_wallet_count is defined %}
+coinjoinmultisession=1
+coinjoinsessions=16
+{% endif %}
```
strophy, line 123: Will this work on testnet if it is under the `[devnet]` section? Either way let's move it up in the file so the logic is clear to someone reading the conf

ansible/group_vars/all
```diff
@@ -102,6 +102,10 @@ dashd_minimumdifficultyblocks: 4032
 # You can provide your own seed node in network config. May be useful if you want to connect to another network
 dashd_seednode: '{{ hostvars[groups.seed_nodes[0]]["public_ip"] if groups.seed_nodes else none }}'
 
+#Do the same for walletnode so we can identify later
+dashd_walletnode: '{{ hostvars[groups.seed_nodes[0]]["public_ip"] if groups.wallet_nodes else none }}'
```
strophy, line 106: Shouldn't this be `hostvars[groups.wallet_nodes[0]]["public_ip"]`? I'm not clear what this line does, can you explicitly state it in the comment, I'm not clear what "the same" is, and when "later"is.

ansible/stage2.yml
```diff
@@ -1,13 +1,32 @@
...
+        restarted: true
+        pull: true
+        timeout: 30
+    
+    - name: Check we're up after restart
+      ansible.builtin.command: dash-cli -rpcwallet=coinjoin-wallet-0 getcoinjoininfo
+      register: coinjoin_output
+      retries: 3
+      delay: 10
+      until: coinjoin_output is not failed
```
strophy, line 28: Is this the same?

```suggestion
      until: coinjoin_output.rc == 0
```

ansible/stage2.yml
```diff
@@ -1,13 +1,32 @@
 ---
-- name: Execute getbestchainlock
+
+#We need to check first to ensure mixing is not enabled from a previous run, if it tries to do it again it will fail (thus not idempotent)
+- name: Check mixing status and start mixing if not enabled
   hosts: wallet_nodes
+  become: true
   tasks:
```
strophy, line 7: Can we move these tasks into the `mixing` role? In `deploy.yml` we only include roles in the top level, which makes handling conditions in stage 2 easier (e.g. when we add proposals next).

Also the step name makes it seem like these tasks will only run in case a condition is satisfied, but there is no condition specified here.

ansible/stage2.yml
```diff
@@ -1,13 +1,32 @@
...
+        timeout: 30
+    
+    - name: Check we're up after restart
+      ansible.builtin.command: dash-cli -rpcwallet=coinjoin-wallet-0 getcoinjoininfo
+      register: coinjoin_output
+      retries: 3
+      delay: 10
+      until: coinjoin_output is not failed
+
+    - name: Include mixing role if coinjoin is not enabled
```
strophy, line 30: Task name implies a condition, but there is no condition here. `mixing` role will run every time, right?

ansible/stage2.yml
```diff
@@ -1,13 +1,32 @@
 ---
-- name: Execute getbestchainlock
+
+#We need to check first to ensure mixing is not enabled from a previous run, if it tries to do it again it will fail (thus not idempotent)
```
strophy, line 3: Where is this check?

ansible/roles/mixing/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+
+- name: Get wallet balance
+  ansible.builtin.command:
+    cmd: dash-cli -rpcwallet=coinjoin-wallet-{{ item | string }} getbalance
+  register: getbalance_result
+  loop: "{{ range(0, coinjoin_wallet_count | int, 1) }}"
+
+- name: Set CoinJoin amount
+  ansible.builtin.command:
+    cmd: dash-cli -rpcwallet=coinjoin-wallet-{{ item | string }} setcoinjoinamount 10
```
strophy, line 34: Let's make the amount a variable to be set in network config

ansible/roles/mixing/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+  loop: "{{ range(0, coinjoin_wallet_count | int, 1) }}"
+
+- name: Set CoinJoin rounds
+  ansible.builtin.command:
+    cmd: dash-cli -rpcwallet=coinjoin-wallet-{{ item | string }} setcoinjoinrounds 16
```
strophy, line 39: Let's make the rounds a variable to be set in network config

ansible/roles/dashd/tasks/main.yml
```diff
@@ -123,14 +123,14 @@
...
-    cmd: dash-cli createwallet {{ wallet_rpc_wallet_faucet }}
+    cmd: dash-cli createwallet {{ wallet_rpc_wallet_faucet }} false false "" false true #Reload wallet on restart
     creates: "{{ dashd_home }}/.dashcore/{{ dash_network_name if dash_network == 'devnet' else 'testnet3' }}/wallets/{{ wallet_rpc_wallet_faucet }}/wallet.dat"
   register: task_result
   when: enable_wallet is true
 
 - name: Create mno wallet
   ansible.builtin.command:
-    cmd: dash-cli createwallet {{ wallet_rpc_wallet_mno }}
+    cmd: dash-cli createwallet {{ wallet_rpc_wallet_mno }}false false "" false true #Reload wallet on restart
```
strophy, line 133:
```suggestion
    cmd: dash-cli createwallet {{ wallet_rpc_wallet_mno }} false false "" false true #Reload wallet on restart
```

strophy (commented):
Good work, a few comments while I'm running a network to test functionality:
- Tasks should be inside a role. You can conditionally include subtasks in a separate yaml file inside the role to conditionally create the wallets, after checking if correct count of wallets already exists - remember to handle the case of increasing `coinjoin_wallet_count` (probably don't worry about decreasing?)
- Looks like more work needed before this is idempotent, some conditions are missing as described in review comments
- What happens when mixing is complete? We need to create an autospend script and load it into crontab as described in the [spec](https://dashpay.slack.com/archives/CRLRS1GBS/p1678976086780489?thread_ts=1678240353.501229&cid=CRLRS1GBS)
- Comments should follow style of space after `#`

strophy:
Also let's add a check for correct spork settings, and presence/functioning of correct quorums to ensure IS locks before we allow mixing to start

ansible/stage2.yml
ktechmidas: Ah this got changed, turns out if we put coinjoinmultisession in the dashd conf it starts automatically and works idempotently since createwallet will simply come back as OK if it's already there, I will remove this comment

ktechmidas:
> Also let's add a check for correct spork settings, and presence/functioning of correct quorums to ensure IS locks before we allow mixing to start

*Technically* sporks and quorums aren't required for privatesend, do we need to do this or can we simply do it for proposals later instead?

ansible/stage2.yml
ktechmidas: Same same, want me to change it to that?

strophy: Yeah, let's keep it consistent with how we check for success in similar commands in the rest of deploy tool

ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -78,6 +78,11 @@ sporkaddr={{ dashd_sporkaddr }}
 sporkkey={{ dashd_sporkkey }}
 {% endif %}
 
+{% if dashd_walletnode is not none and coinjoin_wallet_count is defined %}
```
strophy, line 81: What if `coinjoin_wallet_count` is `0`?

strophy:
Please also add `coinjoin_wallet_count` to the config generator

ansible/deploy.yml
```diff
@@ -200,6 +211,20 @@
...
+  tasks:
+
+    - name: Extract mixer addresses
+      ansible.builtin.set_fact:
+        mixer_addresses: "{{ mixers.values() | map(attribute='owner.address') | list }}"
+    - name: include funding for mixers
+      ansible.builtin.include_tasks: ./roles/mn_fund_collateral/tasks/fund_collateral.yml
+      vars:
+        payment_targets: "{{ mixer_addresses }}"
+        amount: 100
```
strophy, line 226: Why does this need to take place during stage 1?

ansible/roles/dashd/tasks/main.yml
```diff
@@ -123,14 +123,14 @@
 # Wallet node
 - name: Create faucet wallet
   ansible.builtin.command:
-    cmd: dash-cli createwallet {{ wallet_rpc_wallet_faucet }}
+    cmd: dash-cli createwallet {{ wallet_rpc_wallet_faucet }} false false "" false true #Reload wallet on restart
```
strophy, line 126: Why does the wallet need to be reloaded through a `createwallet` command on startup? How is this different to the block on line 101?

ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -78,6 +78,11 @@ sporkaddr={{ dashd_sporkaddr }}
...
+coinjoinmultisession=1
+coinjoinsessions=16
```
strophy, line 83: Maximum is 10, see https://docs.dash.org/projects/core/en/stable/docs/dashcore/wallet-arguments-and-commands-dashd.html#coinjoin-options

ansible/roles/dashd/templates/dash.conf.j2
strophy: This condition still doesn't make sense to me, is `coinjoin_wallet_count` per mixing node?

ansible/roles/dashd/templates/dash.conf.j2
```diff
@@ -78,6 +78,11 @@ sporkaddr={{ dashd_sporkaddr }}
 sporkkey={{ dashd_sporkkey }}
 {% endif %}
 
+{% if dashd_walletnode is not none and coinjoin_wallet_count is defined %}
+coinjoinmultisession=1
```
strophy, line 82: Should we add `coinjoinautostart=1` here too?

ansible/roles/dashd/templates/dash.conf.j2
strophy: I think the condition here should be `dashd_ismixingnode`

ansible/roles/mixing/templates/remix.sh.j2
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+result=$(dash-cli getbalances)
+coinjoin_value=$(echo "$result" | jq '.mine.coinjoin')
+
+# Check if coinjoin_value is greater than 30
+if (( $(echo "$coinjoin_value > 30" | bc -l) )); then
```
strophy, line 7: Let's add logging here so we can easily see when recent transactions were sent

ansible/roles/dashd/tasks/main.yml
ktechmidas: This ensures the wallet will load when dashd is restarted, if you restart dashd on wallet1 without this - it will not load the faucet wallet. 

I can put this in a separate PR since it's technically a bug fix

ansible/roles/dashd/templates/dash.conf.j2
ktechmidas: Would rather not, let's start it manually when we run stage2 so we can control it. We may have a network with coinjoin wallet count that we don't want to start yet

ansible/stage2.yml
```diff
@@ -1,14 +1,8 @@
...
-    - name: Check for chainlock
-      ansible.builtin.fail:
-        msg: "Chainlock failure: no chainlock found."
-      when: '"error" in chainlock_output.stderr'
+- name: Include mixing role
+  hosts: mixer_nodes
+  become: true
+  roles:
+    - role: mixing
+      mnlist: "{{ mixers }}"
```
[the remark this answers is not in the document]
ktechmidas: >     cmd: 'dash-cli -rpcwallet=coinjoin-wallet-0 importprivkey {{ mixers[inventory_hostname].owner.private_key }}'

ansible/roles/dashd/tasks/main.yml
ktechmidas: Either the block on line 101 was added later (and ported in) or doesn't work for some reason. I will test and remove if it's not needed anymore

ansible/stage2.yml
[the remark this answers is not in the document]
strophy: The variable being set is `mnlist` not `mixers`, it does not appear to be used in the `mixing` role, right?

ktechmidas:
@strophy this is now ready for complete testing, and works with any amount of mixers.

* Amount of mixers is specified during the /bin/generate script with -m (or --mixer-nodes) like with seed nodes. If none are given, it will assume 0

* Also during the generation it will add 3 new vars into the .yml file with sensible defaults (wallet name for mixing, amount to mix and the remixing threshold)

terraform/aws/variables.tf
```diff
@@ -207,3 +211,8 @@ variable "create_eip" {
   type        = bool
   default     = true
 }
+
+variable "mixing_enabled" {
```
strophy, line 215: What is this used for?

ansible/roles/dashd/templates/dash.conf.j2
strophy: @ktechmidas this still makes no sense to me

ansible/group_vars/all
```diff
@@ -102,6 +102,10 @@ dashd_minimumdifficultyblocks: 4032
 # You can provide your own seed node in network config. May be useful if you want to connect to another network
 dashd_seednode: '{{ hostvars[groups.seed_nodes[0]]["public_ip"] if groups.seed_nodes else none }}'
 
+# This allows us to identify a wallet node so we can add the mixing stuff in the dashd role
```
strophy, line 105: Are you confusing wallet nodes and mixing nodes here?

ansible/roles/mixing/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+  ansible.builtin.apt:
+    name: jq
+    state: present
+    update_cache: true
+  become: true
+
+- name: Create CoinJoin wallets
+  ansible.builtin.command:
+    cmd: dash-cli createwallet {{ coinjoin_wallet_name }} false false "" false true
+    creates: "{{ dashd_home }}/.dashcore/{{ dash_network_name if dash_network == 'devnet' else 'testnet3' }}/wallets/{{ coinjoin_wallet_name }}/wallet.dat"
```
strophy, line 15: If a mixer node restarts, this step will (correctly) be skipped because the wallet file already exists. Will the wallets still be loaded after the restart?

ansible/roles/mixing/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: Ensure jq is installed
+  ansible.builtin.apt:
+    name: jq
+    state: present
+    update_cache: true
+  become: true
+
+- name: Create CoinJoin wallets
+  ansible.builtin.command:
+    cmd: dash-cli createwallet {{ coinjoin_wallet_name }} false false "" false true
```
strophy, line 14: This causes infinite recursion of the following message:
```
fatal: [mixer-1]: FAILED! => 
  msg: 'An unhandled exception occurred while templating ''{{ coinjoin_wallet_name }}''. Error was a <class ''ansible.errors.AnsibleError''>, original message: An unhandled exception occurred while templating ''{{ coinjoin_wallet_name }}''.
```

lib/configGenerator/generateAnsibleConfig.js
```diff
@@ -114,6 +115,11 @@ async function generateAnsibleConfig(
   // Governance proposals
   config.governance_proposal_count = 2;
 
+  // Mixer variables
+  config.mix_amount = 100;
+  config.remix_amount = 30;
+  config.coinjoin_wallet_name = "{{ coinjoin_wallet_name }}";
```
strophy, line 121: This results in the config YAML file containing a Jinja2 template like `coinjoin_wallet_name: '{{ coinjoin_wallet_name }}'`, but this file is not templated. How is this supposed to work? Why do we need to be able to configure the wallet name?

ansible/roles/mixing/tasks/main.yml
ktechmidas: outdated

terraform/aws/variables.tf
ktechmidas: Removed

ansible/group_vars/all
ktechmidas: No, this was when we were using the wallet nodes and mixing nodes weren't a thing.... that approach completely failed. I'll update it

ansible/roles/mixing/tasks/main.yml
ktechmidas: Yes due to the "true" at the end, it's "load_on_startup"

https://docs.dash.org/projects/core/en/stable/docs/api/remote-procedure-calls-wallet.html#createwallet

ansible/roles/mixing/tasks/main.yml
ktechmidas: Is that infinite, or do you just not have coinjoin_wallet_name in your .yml file? Can you check?

lib/configGenerator/generateAnsibleConfig.js
ktechmidas: Woah okay that answers my previous question... that should be an actual value... :)

I think a search and replace probably went wrong, fixed

ansible/roles/dashd/templates/dash.conf.j2
ktechmidas: As mentioned elsewhere, this whole code is now useless and needs removing...

1) We set the variables on startup of mixing, not in the config file

2) It used to be the wallet node, we had 3 mixing wallets on a wallet node (your initial implementation) which didn't work, so went to using mixing nodes. 

I will remove it

ansible/roles/mixing/tasks/main.yml
strophy: @ktechmidas fixed in https://github.com/dashpay/dash-network-deploy/pull/528/commits/508d6866de003408600933fd3a78133a51e38c0e

ansible/roles/dashd/tasks/main.yml
strophy: @ktechmidas you added the 101 block yourself in https://github.com/dashpay/dash-network-deploy/pull/438

Did you test this? I haven't seen any evidence it doesn't work, and the RPC command will just edit `dash.conf` for us, so I'm going to remove this and rely on Ansible directly for templating instead of templating by proxy through dashd.

ansible/roles/mixing/templates/remix.sh.j2
strophy: @ktechmidas I attempted this in https://github.com/dashpay/dash-network-deploy/pull/528/commits/44135d223ccae0853b3fdd6791770d1895ee8098

strophy (approved):
Well done mate! :partying_face:

Issue #514, opened 2023-06-15
Remove mn_evo_services role from seed nodes


pshenmic:
## Expected Behavior

Seed nodes should only run core tenderdash and seed containers, without dapi, envoy and the rest of the stuff

## Current Behavior

Legacy mn_evo_services role is used to deploy all platform services on the host

shumkov:
Actually we need dashd there, but we don't need mn_evo_services except Tenderdash

pshenmic:
Sorry guys, I think I messed up dashd with mn_evo_services role in the beginning and confused you

pshenmic:
https://github.com/dashpay/dash-network-deploy/pull/535

PR #535, opened 2023-08-12
feat: dashmate for seeds

pshenmic:

https://github.com/dashpay/dash-network-deploy/issues/514

## What was done?

* Removed mn_evo_services role
* Stop unused containers at the end of dashmate role for seed nodes

## How Has This Been Tested?

On my own devnet + smoke tests

## Breaking Changes

No

## Checklist:


strophy:
I'll try and deploy this on testnet seed-3 once the 0.24.22 dashmate release is complete.

strophy (approved):
Code reviewed and approved in concept, waiting for dashmate release for testing on testnet, so please don't merge yet.

strophy:
There appears to be an error in the logic on initial startup, where immediately after starting all services, the dashmate role then attempts to restart them, which fails. Ansible output:
```
TASK [dashmate : Print status] *******************************************************************************************************************************
ok: [seed-1] => 
  msg: |2-
  
     Core Started: False
     Platform Started: False
     Core Images Updated: False
     Platform Images Updated: True
     Dashmate Package Updated: True
     Dashmate Config Changed: True

TASK [dashmate : Start dashmate services] ********************************************************************************************************************
changed: [seed-1]

TASK [dashmate : Start dashmate platform services] ***********************************************************************************************************
skipping: [seed-1]

TASK [dashmate : Restart dashmate services] ******************************************************************************************************************
fatal: [seed-1]: FAILED! => 
  msg: |-
    The conditional check 'dashmate_start_platform.skipped and dashmate_start.skipped' failed. The error was: error while evaluating conditional (dashmate_start_platform.skipped and dashmate_start.skipped): 'dict object' has no attribute 'skipped'. 'dict object' has no attribute 'skipped'
  
    The error appears to be in '/home/strophy/dash-network-deploy/ansible/roles/dashmate/tasks/main.yml': line 239, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
    - name: Restart dashmate services
      ^ here
```

strophy:
This is also not quite idempotent yet, running the `dashmate` role twice on a seed node will always result in `is_dashmate_config_changed` being true, and all processes are restarted even though nothing is changed. We need to look at the `ansible.builtin.template` command on line 113 and figure out why this always reports status `changed`.

strophy:
Other than the above logic issue, this seems to be working fine. It's a bit inefficient to run through the whole ZeroSSL workflow, assign a cert, and then shut down the DAPI container while leaving all the provisioned config and containers in place, but I think changing this would require significant changes to dashmate right? I think we can leave it for now, it's really great to be able to remove the entire `mn_evo_services` role!

strophy:
Please add `mode: "validator"` to EN dashmate invocation in `deploy.yml` after line 238 to avoid this error:
```
TASK [dashmate : Write dashmate config file] *******************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'mode' is undefined. 'mode' is undefined
fatal: [hp-masternode-1]: FAILED! => changed=false 
  msg: 'AnsibleUndefinedVariable: ''mode'' is undefined. ''mode'' is undefined'
```

strophy:
While trying to update evonodes to the v0.24.23 in order to fix issues related to restarts and log files on testnet, I found that dashmate fails to stop containers on evonodes:
```
dashmate@hp-masternode-1:~$ dashmate stop
✔ Check node is running
✔ Stopping testnet node
dashmate@hp-masternode-1:~$ docker ps
CONTAINER ID   IMAGE                       COMMAND                  CREATED       STATUS      PORTS                                                                                                        NAMES
2db3a4fd2b3b   dashpay/tenderdash:0.11.3   "docker-entrypoint.s…"   2 weeks ago   Up 4 days   26656-26657/tcp, 127.0.0.1:6060->6060/tcp, 127.0.0.1:36657->36657/tcp, 26660/tcp, 0.0.0.0:36656->36656/tcp   dashmate_testnet-drive_tenderdash-1
59a46d4fee19   dashpay/drive:0.24.19       "docker-entrypoint.s…"   2 weeks ago   Up 4 days   26658/tcp                                                                                                    dashmate_testnet-drive_abci-1
c5bb65518a3c   dashpay/dashd:19.3.0        "/docker-entrypoint.…"   2 weeks ago   Up 4 days   9998-9999/tcp, 127.0.0.1:19998->19998/tcp, 0.0.0.0:19999->19999/tcp                                          dashmate_testnet-core-1
dashmate@hp-masternode-1:~$ dashmate --version
dashmate/0.24.23 linux-x64 node-v16.20.1
```

pshenmic:
> dashmate_start_platform.skipped and dashmate_start.skipped
Weirdly, I remember fixing this already in this branch 🤔 . Fixed once again

> This is also not quite idempotent yet, running the dashmate role twice on a seed node will always result in is_dashmate_config_changed being true
Yes, the idempotency is not yet stable, because we often change a lot of values during dashmate setup manually, that causes config re-render.
To be honest, I feel like restart logic that we came up with is too complex to be stable

> Other than the above logic issue, this seems to be working fine. It's a bit inefficient to run through the whole ZeroSSL workflow, assign a cert, and then shut down the DAPI container while leaving all the provisioned config and containers in place, but I think changing this would require significant changes to dashmate right? I think we can leave it for now, it's really great to be able to remove the entire mn_evo_services role.

> 'AnsibleUndefinedVariable: ''mode'' is undefined
Fixed

> dashmate fails to stop all the containers on evonodes:
I am going to look into that now

strophy:
> Though we will still need dapis on the seeds later.

In that case we can just leave the current setup, and change behaviour around the DAPI container shutdown when we enable it later.

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -269,6 +269,35 @@ Dashmate Config Changed: {{ is_dashmate_config_changed }}\n"
...
+      }}
+  when: mode == 'seed'
+
+- name: Stop some services in seed mode
+  community.docker.docker_container:
+    name: "{{ item }}"
+    state: stopped
+  when: mode == 'seed'
+  loop: "{{ containers_to_stop }}"
+
```
shumkov, line 300: What are we doing here? Task names are not representative. Please rename tasks and add comments.

ansible/roles/dashmate/tasks/main.yml
```diff
@@ -269,6 +269,38 @@ Dashmate Config Changed: {{ is_dashmate_config_changed }}\n"
     chdir: '{{ dashmate_cwd }}'
   changed_when: build_dashmate_helper.rc == 0
 
+# In the seed mode, we are disabling some services that are not necessary to run on the node
```
shumkov, line 272: I believe DAPI and envoy are still required. DAPI Client connects to a seed node to get SML through DAPI. So only Drive is actually not needed on seed nodes but it's ok if it's running - it won't do anything and consequently won't produce any load.

shumkov (changes requested):

ansible/roles/dashmate/tasks/main.yml
pshenmic: Ok, but I wonder how does it work now? DAPI on the seeds is not working for a long time in the mn_evo_services

strophy: What is needed to get it working? Should we investigate that in this PR, or separately?

strophy: DAPI currently fails as follows on seed nodes:
Dashmate node:
```
  1) Smoke
       DAPI
         All nodes
           seed-1
             should return data from Core:
     AssertionError: FetchError: invalid json response body at https://35.92.167.154:1443/ reason: Unexpected token u in JSON at position 0
      at Context.<anonymous> (test/smoke/dapi.js:79:20)
```
```
  [Symbol(Response internals)]: {
    url: 'https://35.92.167.154:1443/',
    status: 503,
    statusText: 'Service Unavailable',
    headers: Headers { [Symbol(map)]: [Object: null prototype] },
    counter: 0
  }
```
Non-working mn_evo_services node:
```
  7) Smoke
       DAPI
         All nodes
           seed-3
             should return data from Core:
     AssertionError: ResponseError: request to https://18.246.62.131:1443/ failed, reason: write EPROTO 140119617955776:error:1408F10B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:332:

      at Context.<anonymous> (test/smoke/dapi.js:79:20)
```

ktechmidas (approved):
utACK

ansible/roles/dashmate/tasks/main.yml
strophy: This will be resolved in a separate PR, since adding DAPI functionality to seed nodes is out of scope for this PR

test/smoke/tenderdash.js
```diff
@@ -62,7 +62,6 @@ describe('Tenderdash', () => {
 
           if (error) {
             // noinspection ExceptionCaughtLocallyJS
-            throw new Error(error.message);
```
strophy, line 65: Why remove this?

strophy (approved):

PR #539, opened 2023-08-30
chore: speedup address import

strophy:

Waiting for a rescan of the entire wallet each time nodes needed unbanning was tedious and usually unnecessary in the first place.

## What was done?

We now check if the addresses we are about to import are already imported before importing them. This still takes the amount of time needed to `mn_count * 2` RPC commands, but we can skip the subsequent rescan step, saving a bit of time on testnet unban procedures.

## How Has This Been Tested?

On testnet 

## Breaking Changes

None

## Checklist:


strophy:
Can anyone think of a better way to check if the addresses are already in the wallet?

ktechmidas (approved):

PR #541, opened 2023-09-04
chore: support Platform v0.25

strophy:

Numerous updates were needed to deploy using dashmate v0.25, and run tests against v0.25 networks

## What was done?

- pin wasm-bindgen version
- bump node version to 18
- handle zerossl logic error on first run when config is not yet set
- remove sentinel from dashmate config
- bump deps and version
- fix linter errors

## How Has This Been Tested?

On devnet-absinthe and devnet-ouzo

## Checklist:

### DIFF
--- a/ansible/roles/dashmate/tasks/build.yml
+++ b/ansible/roles/dashmate/tasks/build.yml
@@ -94,6 +94,8 @@
 
 - name: Check if default config is set
   ansible.builtin.command: "{{ dashmate_cmd }} config"
+  become: true
+  become_user: dashmate
   register: dashmate_config_result
   ignore_errors: true
   changed_when: true
--- a/ansible/roles/dashmate/tasks/build.yml
+++ b/ansible/roles/dashmate/tasks/build.yml
@@ -49,7 +49,7 @@
   ansible.builtin.include_role:
     name: geerlingguy.nodejs
   vars:
-    nodejs_version: "16.x"
+    nodejs_version: "18.x"
 
 - name: Clone dashmate source
   become: true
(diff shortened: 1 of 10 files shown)


ansible/roles/dashmate/tasks/main.yml
```diff
@@ -107,6 +107,16 @@
   ansible.builtin.set_fact:
     dashmate_config_version: "{{ (dashmate_version_result.stdout | split(' ') | first() | split('/'))[1] }}"
 
+- name: Check if default config is set
```
shumkov, line 110: Why do we need this?

ansible/roles/dashmate/tasks/build.yml
```diff
@@ -92,6 +92,16 @@
   when: git.changed # noqa: no-handler
   changed_when: true
 
+- name: Check if default config is set
```
shumkov, line 95: Why do we need this?

strophy: Because immediately after, we run `dashmate config set`. This command will fail on the initial run of deploy tool, because the default config is not yet set.

ansible/roles/dashmate/tasks/main.yml
strophy: Because immediately after, we run `dashmate config get`. This command will fail on the initial run of deploy tool, because the default config is not yet set.

ansible/roles/dashmate/tasks/build.yml
shumkov: You mean config is not rendered yet?

strophy: I mean `defaultConfigName` is not yet set, so commands like `dashmate config set ...` cannot work

shumkov (approved):
👍

PR #577, opened 2023-10-11
fix: use tenderdash instead of drive var

ktechmidas:
drive_host_info var no longer exists

## What was done?
Use tenderdash_host_info instead

## Checklist:

### DIFF
--- a/ansible/roles/elastic_beats/vars/platform.yml
+++ b/ansible/roles/elastic_beats/vars/platform.yml
@@ -2,7 +2,7 @@
 
 filebeat_platform_inputs:
   - type: log
-    enabled: "{{ drive_host_info.containers | length > 0 }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/elastic_beats/vars/platform.yml
```diff
@@ -2,7 +2,7 @@
 
 filebeat_platform_inputs:
   - type: log
-    enabled: "{{ drive_host_info.containers | length > 0 }}"
+    enabled: "{{ tenderdash_host_info.containers | length > 0 }}"
```
shumkov, line 5: Not sure it's a good idea because tenderdash is running on seed nodes but without Drive

shumkov (changes requested):

strophy:
Let's go with #581 instead

PR #607, opened 2023-12-13
feat: metric and load test nodes

shumkov:

We need to run load tests on the network and monitor how Platform behaves 
 

## What was done?

- Added metric nodes
- Added load test nodes
- Configured tenderdash mempool cache

## How Has This Been Tested?

Deploying networks

## Breaking Changes

Platform < 1.0.0-dev.4 is not supported

## Checklist:


bin/generate-configs.js
```diff
@@ -38,6 +40,7 @@ async function main() {
       hpMasternodesAmdCount,
       hpMasternodesArmCount,
       mixerCount,
+      loadTestCount,
```
shumkov: Generate config doesn't set metrics (let's set always 1 into config) and load test nodes count

terraform/aws/instances.tf
```diff
@@ -454,6 +454,80 @@ resource "aws_instance" "logs" {
     DashNetwork = terraform.workspace
   }
 
+  lifecycle {
+    ignore_changes = [ami]
```
shumkov, line 458: I guess we should remove it

terraform/aws/variables.tf
```diff
@@ -56,6 +56,11 @@ variable "tendermint_p2p_port" {
   default     = 36656
 }
 
+variable "prometheus_port" {
+  description = "Prometheus port"
+  default     = 80
```
shumkov, line 61: Is it actually prometheus default port?

terraform/aws/instances.tf
ktechmidas: Oh yeah looks like a copy paste error on my part...

ktechmidas: Wait... why are we removing this? It's attached to the logs section, we haven't touched it in this PR?

ktechmidas: Removing this part would cause terraform to _destroy_ and recreate the instance when it's run and the AMI has been updated. Ideally we don't want it to do that on any servers with a running network

bin/generate-configs.js
ktechmidas: Fixed, I've removed them from /bin/generate because it's getting too complicated. If someone wants to change they can change the tfvars

terraform/aws/variables.tf
ktechmidas: Default is 9090, we set 80 so it's easier to access. Let's not change this

Our default and Prometheus default are different

ktechmidas (dismissed):

ansible/roles/protobuf_compiler/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    path: "{{ protoc_dest_path }}"
+    state: directory
+    mode: '0755'
+
+- name: Unzip protoc to the destination
+  ansible.builtin.unarchive:
+    src: "{{ protoc_zip_path }}/{{ protoc_zip_file_name }}"
+    dest: "{{ protoc_dest_path }}"
+    remote_src: true
+# TODO: Create an empty file with version and then check if it's exists
```
shumkov, line 18: Create a file with current version in name after installation of protoc. If file exists we can skip all these tasks

bin/generate
```diff
@@ -42,17 +42,13 @@ if [[ -z ${HP_MASTERNODES_AMD_COUNT} ]] || [[ -z ${HP_MASTERNODES_ARM_COUNT} ]];
 fi
 
 SEEDS_COUNT="1"
-MIXER_COUNT="0"
```
shumkov, line 45: Why did we remove mixer count here?

lib/cli/terraform.sh
```diff
@@ -9,7 +9,7 @@ function terraform_init() {
 
     # Initialize terraform w/remote AWS backend configuration
 
-    terraform init \
+    terraform init -reconfigure \
```
shumkov: That's correct?

lib/configGenerator/generateAnsibleConfig.js
```diff
@@ -42,6 +42,7 @@ async function generateAnsibleConfig(
   hpMasternodesCount,
   seedsCount,
   mixerCount,
+  loadTestCount,
```
shumkov: Not sure if it's actually used

bin/generate
```diff
@@ -61,4 +57,4 @@ done
 node ./bin/generate-configs.js ${NETWORK} ${NETWORK_NAME} \
   ${MASTERNODES_AMD_COUNT}    ${MASTERNODES_ARM_COUNT}    \
   ${HP_MASTERNODES_AMD_COUNT} ${HP_MASTERNODES_ARM_COUNT} \
-  ${SEEDS_COUNT} ${MIXER_COUNT}
```
shumkov, line 64: I believe we should have LOAD_TEST_COUNT here

lib/cli/terraform.sh
ktechmidas: Fixed

bin/generate
ktechmidas: As discussed, generate is being made easier to use. Another PR will go in after this one to pull values from tfvars

lib/configGenerator/generateTerraformConfig.js
```diff
@@ -7,12 +7,13 @@ async function generateTerraformConfig(
...
   hpMasternodesArmCount,
-  mixerCount,
 ) {
   const config = {
     miner_count: 1,
     logs_count: 1,
-    mixer_count: mixerCount,
+    mixer_count: 0,
+    load_test_count: 1,
+    metrics_count: 1,
```
shumkov: I would say those two must be 0 by default

lib/configGenerator/generateAnsibleConfig.js
```diff
@@ -220,10 +220,17 @@ async function generateAnsibleConfig(
...
 
-  for (let i = 1; i <= mixerCount; i++) {
+  for (let i = 1; i <= 0; i++) {
     config.mixers[`mixer-${i}`] = { owner: await generateDashAddress(network) };
   }
 
+  // Load tools
+  config.load_testers = {};
+
+  for (let i = 1; i <= 1; i++) {
```
shumkov: I would say it should be zero by default as well

QuantumExplorer:
Please fix linting issues before merge.

ktechmidas:
@QuantumExplorer linting issues fixed, going to merge

PR #608, opened 2023-12-20
chore!: update dashmate config for v0.25.20

shumkov:

We need to update dashmate config to represent changes in v0.25.20

## What was done?

- Added new configuration options and corresponding variables
 

## How Has This Been Tested?

None

## Breaking Changes

Dashmate < 0.25.20 is not supported

## Checklist:


ansible/roles/dashmate/templates/dashmate.json.j2
```diff
@@ -213,7 +224,25 @@
             },
             "consensus": {
               "createEmptyBlocks": true,
-              "createEmptyBlocksInterval": "3m"
+              "createEmptyBlocksInterval": "3m",
+              "peer": {
+                "gossipSleepDuration": "{{dashmate_platform_drive_tenderdash_consensus_peer_gossip_sleep_duration}}",
+                "queryMaj23SleepDuration": "{{dashmate_platform_drive_tenderdash_consensus_peer_query_maj23_sleep_duration}}"
+              },
+              "unsafeOverride": {
```
lklimek, line 232: Why not using consensus params instead of overrides?

shumkov: We can't change the consensus param without resetting the network. Having these options we can experiment with numbers quickly. It's not for persistent configuration.

QuantumExplorer (approved):

ansible/roles/dashmate/templates/dashmate.json.j2
QuantumExplorer: Acually we can chance the consensus params once an hour on testnet.

QuantumExplorer: But I agree this is faster for now.

shumkov: You mean if we hardcode this change versions?

PR #637, opened 2024-04-29
fix: faucet role is not compatible with docker compose v2

ktechmidas:
Fixed some faucet options for docker compose v2 + added a promo code for "platform" so people can get extra tDash

## What was done?
Changed options to standard
Added promo code

## How Has This Been Tested?
Testnet

## Breaking Changes
None

## Checklist:


ansible/roles/multifaucet/tasks/main.yml
```diff
@@ -45,24 +45,32 @@
...
     command: >
       mysql -h localhost -D multifaucet -u multifaucet --password="multifaucet"
       -e 'CREATE TABLE IF NOT EXISTS faucet_promo_codes (id int NOT NULL, code varchar(255), minimum_payout int, maximum_payout int, PRIMARY KEY (id));'
 
 - name: Load MN promo code
   community.docker.docker_container_exec:
-    container: multifaucet_db_1
+    container: multifaucet-db-1
     command: >
       mysql -h localhost -D multifaucet -u multifaucet --password="multifaucet"
```
shumkov, line 57: You can pass all three rows after VALUES instead of calling client 3 times

shumkov (changes requested):

ansible/roles/multifaucet/tasks/main.yml
ktechmidas: One of them only applies to devnets, while it's wasting a few CPU cycles it won't actually have any effect since there's no actual traffic at this point in setup

shumkov (approved):

PR #641, opened 2024-05-08
Feat: Change in Tenderdash Role to investigate the chain halt issue.

vivekgsharma:
## Issue Being Fixed or Feature Implemented

1. Configured log rotation for Tenderdash logs
2. Added a Logstash filter to exclude TRACE logs from processing to optimize log
3. Updated the `tenderdash_log_level` to "trace" in the Ansible variables 

## Changes Made

- Added a logrotate configuration for Tenderdash logs to ensure logs are rotated daily, compressed, and old logs are removed after three days.
- Implemented a Logstash configuration to drop TRACE level logs to prevent them from being processed and stored
- Set the `tenderdash_log_level` variable to "trace"  in the ansible Varibales

## Breaking Changes

No breaking changes introduced.

## Checklist:
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.


vivekgsharma:
Test is failing due to `Risky file permission`  due to this line ` create 0640 tenderdash adm ` in the new added task , this permission might be necessary to access the log files. I can ignore this check after the review adding ignore lines in .ansible-lint

ktechmidas:
@vivekgsharma what is adm? can't it use 600?

vivekgsharma:
> @vivekgsharma what is adm? can't it use 600?

`adm` is a system group, members of this group can read logs in various directories (/var/log). We can try 600 

Reference for using 0640 
https://serverfault.com/questions/694707/what-should-be-the-right-logs-permissions-for-nginx-on-centos

vivekgsharma:
@ktechmidas changed to 0644 because most of the tasks are using this mode in the current setup. Ansible-lint checks have passed

ansible/roles/tenderdash/tasks/main.yml
```diff
@@ -50,3 +50,32 @@
...
+    dest: /etc/logrotate.d/tenderdash
+    content: |
+      {{ tenderdash_compose_path }}/tenderdash/logs/*.log {
+          daily
+          rotate 3
+          compress
+          delaycompress
+          missingok
+          notifempty
+          create 0644 {{ tenderdash_uid }} {{ tenderdash_gid }}
```
lklimek: why `0644` and not `0640`? World-readable logs are a potential security risk  (but sometimes it's needed, eg. when Docker needs to use it - please double-check).

ansible/roles/tenderdash/tasks/main.yml
```diff
@@ -50,3 +50,32 @@
...
+    group: '{{ tenderdash_gid }}'
+    mode: '0644'
+  when: tenderdash_network == 'testnet'
+
+- name: Add Logstash filter for Tenderdash TRACE logs
+  ansible.builtin.lineinfile:
+    path: /etc/logstash/conf.d/tenderdash.conf
+    line: |
+      filter {
+        if [log_level] == "TRACE" {
```
lklimek: In Tenderdash, it is lowercase (`trace`)

ansible/roles/tenderdash/tasks/main.yml
```diff
@@ -50,3 +50,32 @@
...
+          compress
+          delaycompress
+          missingok
+          notifempty
+          create 0644 {{ tenderdash_uid }} {{ tenderdash_gid }}
+      }
+    owner: '{{ tenderdash_uid }}'
+    group: '{{ tenderdash_gid }}'
+    mode: '0644'
+  when: tenderdash_network == 'testnet'
```
lklimek: I think we can use this config for all networks, given that logs are saved to elasticsearch.

ansible/roles/tenderdash/tasks/main.yml
```diff
@@ -50,3 +50,32 @@
...
     pull: always
     recreate: always
+
+- name: Configure logrotate for Tenderdash logs
+  ansible.builtin.copy:
+    dest: /etc/logrotate.d/tenderdash
+    content: |
+      {{ tenderdash_compose_path }}/tenderdash/logs/*.log {
+          daily
+          rotate 3
```
lklimek: maybe `rotate 1`

lklimek (changes requested):

ansible/roles/tenderdash/tasks/main.yml
vivekgsharma: Done

ansible/roles/tenderdash/tasks/main.yml
vivekgsharma: Done

ansible/roles/tenderdash/tasks/main.yml
vivekgsharma: fixed

ktechmidas:
Closed - role no longer in use

PR #660, opened 2024-08-12
Dashmate quick update

ktechmidas:
This will run a quick update if you set the var, only set the var if the network is already deployed and just needs updating

## What was done?
Added var for dashmate_quick_update and quick_update.yml with the basics

## Breaking Changes
None

## Checklist:


shumkov (changes requested):
I see  a lot of code duplication. It will be hard to maintain so you need to change same things in both places. I didn't get what is the actual difference. Also I think tags a more appropriate in this case.

PR #738, opened 2026-03-30
feat: add CI automation for devnet create and destroy workflows

vivekgsharma:
Create Devnet using GitHub Actions. The main problems were CI environment drift and a few runtime issues that only showed up in clean CI runs, especially during masternode funding and HP node Dashmate setup.

  ## What was done?

  - Pinned CI to the working Ansible environment used by manual deploys:
      - ubuntu-22.04
      - ansible-core==2.16.3
      - jmespath
  - Fixed config generation issues used by CI-created networks:
      - set Terraform main_domain
      - ensured dashmate_core_rpc_quorum_list_password is generated and rendered correctly
  - Hardened spork activation:
      - safe handling when EHF/spork keys are absent
      - avoided failures on unsupported sporks
  - Updated Dashmate Ansible role behavior for HP nodes:
      - run commands with correct HOME using become_flags: -H
      - pass explicit Dashmate config selection during render/start/update/status/SSL operations
      - retry transient Let’s Encrypt failures
  - Fixed masternode funding flow in Ansible:
      - kept 1-Dash fee funding batched
      - changed 4000-Dash HPMN collateral funding to one-by-one with confirmation between sends
      - this avoids Transaction too large failures from oversized wallet transactions in CI

  ## How Has This Been Tested?

  - Repeatedly tested with GitHub Actions Create Devnet / Destroy Devnet
  - Verified resumed debugging runs against failing Ansible sections during iteration
  - Completed a successful CI run through:
      - masternode registration
      - HPMN collateral funding
      - ProTx creation
      - HP node Dashmate/platform setup
      - post-deploy service verification

  ## Breaking Changes

  None.

  ## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in deployment workflows with safer configuration validation and fallback mechanisms.
  * Enhanced retry logic for SSL certificate operations with configurable timeouts.

* **Chores**
  * Updated deployment infrastructure with strengthened devnet name validation.
  * Optimized deployment process to support resuming interrupted operations.
  * Adjusted configuration defaults for improved system initialization.


.github/workflows/create-devnet.yml
```diff
@@ -52,7 +57,7 @@ on:
 jobs:
   create:
     name: Create Devnet
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ktechmidas, line 60: Does it have an error on latest?

.github/workflows/create-devnet.yml
```diff
@@ -112,7 +117,7 @@ jobs:
       - name: Install Ansible
         run: |
           python3 -m pip install --upgrade pip
-          python3 -m pip install ansible
+          python3 -m pip install ansible-core==2.16.3 jmespath
```
ktechmidas, line 129: Maybe we'll deal this in a separate PR, I know 2.19 broke a lot of stuff and our entire codebase probably needs an update...

.github/workflows/create-devnet.yml
```diff
@@ -124,6 +129,7 @@ jobs:
         env:
           DEPLOY_SERVER_KEY: ${{ secrets.DEPLOY_SERVER_KEY }}
           EVO_APP_DEPLOY_KEY: ${{ secrets.EVO_APP_DEPLOY_KEY }}
+          EVO_APP_DEPLOY_WRITE_KEY: ${{ secrets.EVO_APP_DEPLOY_WRITE_KEY }}
```
ktechmidas, line 141: It would be nice if it would generate a new key for each devnet, maybe we can use GPG to encrypt it and show it? Then you & i can decrypt on our sides and get a new key

ansible/deploy.yml
```diff
@@ -1,342 +1,6 @@
 ---
 
-# Bootstrap
```
ktechmidas, line 3: Should this have been deleted? We still want to be able to run manually if we need to...

ansible/roles/activate_dashd_sporks/tasks/main.yml
```diff
@@ -37,4 +37,6 @@
   vars:
     generate: spork_activation
     spork_name: SPORK_24_TEST_EHF
-  when: not (active_sporks.stdout | from_json)["SPORK_24_TEST_EHF"]
+  when:
+    - "'SPORK_24_TEST_EHF' in (active_sporks.stdout | from_json)"
```
ktechmidas: Actually don't think this exists anymore, we can remove it completely

ktechmidas (commented):
First pass at a review

.github/workflows/create-devnet.yml
vivekgsharma: Yes, on ubuntu-latest I hit CI-only environment drift issues. Pinning to ubuntu-22.04 was to stabilize the workflow against the known-good deploy environment.

.github/workflows/create-devnet.yml
vivekgsharma: The workflow was failing in CI with the newer Ansible from ubuntu-latest/pip install ansible, while the same deploy flow was working manually with Ansible 2.16.3. So this was necessary to remove CI environment drift and get Create Devnet run successfully

ansible/deploy.yml
vivekgsharma: Agreed. Restored it.

ansible/roles/activate_dashd_sporks/tasks/main.yml
vivekgsharma: Agreed, removed the obsolete SPORK_24_TEST_EHF handling.

.github/workflows/create-devnet.yml
vivekgsharma: Understood. I am iterating with the testing and will get this added once you do the final review and PR is about to merge , this is will me in fast testing.

dashinfraclaw (commented):
## Code Review — @dashinfraclaw

Went through the full diff. This is solid CI hardening work. A few observations:

### ✅ Good stuff

- **Pinning `ubuntu-22.04` + `ansible-core==2.16.3`** — essential for reproducibility. `ubuntu-latest` drifting is a classic CI footgun.
- **`become_flags: -H`** on all dashmate tasks — correct fix. Without it, `HOME` stays as root's when using `become_user: dashmate`, which breaks npm/yarn/cargo that write to `$HOME`. This was probably causing intermittent CI failures.
- **Resume mode for existing configs** — smart. Lets you re-run Ansible without re-terraforming when debugging mid-flow failures. The partial config detection (fail if 1 or 2 of 3 files exist) is a nice safety check.
- **Jinja `.get(item, false)` for spork lookups** — fixes KeyError when a spork doesn't exist in the response. The old `[item]` would blow up on missing keys.
- **`default(..., true)` for quorum_list_password** — the `true` param makes Jinja treat empty strings as undefined too, not just `None`. Good catch.
- **One-by-one HPMN collateral funding** — correct fix for "Transaction too large" from batching 4000 DASH × N sends. Regular MN fee funding (1 DASH each) stays batched via `sendmany`, which makes sense since those are small.
- **Let's Encrypt retry (3x, 20s delay)** — rate limits and transient DNS failures are real. Good addition.
- **Write key separation** for configs repo push — clean approach with graceful fallback.

### ⚠️ Minor concerns

1. **EHF spork activation removed entirely** (activate_dashd_sporks/main.yml lines deleted). Was SPORK_24_TEST_EHF intentionally dropped from all devnet flows, or just from the initial spork activation? If EHF is still needed later in the pipeline, this could silently break it.

2. **`fund_one_collateral.yml` uses `sendtoaddress`** — this sends one tx per node and waits for confirmation each time. For 20+ HPMNs that's 20+ block-generation cycles. The PR description acknowledges this is intentional to avoid "Transaction too large", but worth noting it will significantly slow down HPMN-heavy devnet creation. Could be optimized later with small batches (e.g., 5 at a time) instead of strictly one-by-one.

3. **`mn_names` fact construction** — the new dict2items/map pattern is safer than iterating a dict directly, but the `else []` branch means a misconfigured `mnlist` would silently produce zero masternodes instead of failing. Might want a `fail` task if `mn_names` is empty and `mnlist` was expected.

4. **Config push step has `if: always()`** — this means it runs even if the deploy failed. That's probably intentional (save partial progress), but could push incomplete/broken configs to the repo on a mid-deploy failure. The resume mode would then pick up those broken configs on retry. Consider gating on at least Terraform success.

5. **`TF_CLI_ARGS_destroy: "-auto-approve"`** — fine for CI, just noting this means the destroy workflow has no confirmation step at all. The concurrency lock helps, but a manual approval gate on the destroy workflow might be worth considering.

### Nit

- The commit to `generateAnsibleConfig.js` sets `quorum_list_password = dashmate_password` — this means they're always identical. If that's the intent, the Jinja template default is redundant (but harmless as belt-and-suspenders).

Overall: good to go with the minor considerations above. The `become_flags: -H` fix alone is worth merging — that's been a known pain point.

dashinfraclaw:
## 🚨 Destroy Workflow — Testnet/Mainnet Protection Audit

Traced the full destroy path from the GitHub Actions workflow through  →  → . Here's the safety analysis:

### Current protections

1. **Naming prefix:** The workflow hardcodes , so any input gets the  prefix. Entering `testnet` becomes `devnet-testnet`, not `testnet`. ✅
2. **Input validation:** Rejects empty names and names starting with `devnet-` (to prevent `devnet-devnet-...`). ✅
3. **Config file gating:** The workflow copies configs from `dash-network-configs` and fails if the `.yml/.tfvars/.inventory` triple doesn't exist. There's no `devnet-testnet.*` config set, so it would fail. ✅
4. **Terraform workspace isolation:** `init.sh` parses the network name and selects a Terraform workspace matching it. `devnet-*` networks get their own workspace, separate from `testnet`/`mainnet`. ✅
5. **`bin/destroy` has a testnet safeguard:** If `NETWORK_NAME == "testnet"`, it requires typing a solemn oath. But this can't trigger from the workflow since the name always has `devnet-` prefix. ✅

### 🟡 Gaps that should be addressed

**1. No input sanitization on special characters**

The devnet_name input has zero character validation. Someone could enter:
- Spaces, quotes, backticks → potential shell injection in the `./bin/destroy "$NETWORK_NAME"` call
- Very long strings → potential issues with Terraform workspace names or S3 keys

**Suggested fix** — add to the `Validate devnet name` step:
```bash
if [[ ! "$NAME" =~ ^[a-z0-9]([a-z0-9-]*[a-z0-9])?$ ]]; then
  echo "Error: devnet_name must be lowercase alphanumeric with hyphens only"
  exit 1
fi
if [[ ${#NAME} -gt 32 ]]; then
  echo "Error: devnet_name must be 32 characters or less"
  exit 1
fi
```

**2. No explicit blocklist for protected network names**

While the `devnet-` prefix prevents direct `testnet`/`mainnet` targeting, a belt-and-suspenders blocklist would make the intent explicit and survive refactors:

```bash
PROTECTED_NAMES=("testnet" "mainnet" "mainnet-support" "regtest")
for protected in "${PROTECTED_NAMES[@]}"; do
  if [[ "$NAME" == "$protected" || "devnet-$NAME" == "$protected" ]]; then
    echo "Error:  is a protected network name and cannot be used as a devnet name"
    exit 1
  fi
done
```

**3. `TF_CLI_ARGS_destroy: "-auto-approve"` with no manual approval gate**

The destroy workflow runs immediately on dispatch with no GitHub environment protection or approval step. Anyone with `workflow_dispatch` permission can destroy any devnet instantly. Consider:
- Adding a GitHub Environment with required reviewers for the destroy workflow
- Or at minimum, add a summary step with a pause/confirmation before actual destruction

**4. The AWS credentials in secrets have no scope restriction**

The same `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` are used for both create and destroy. If these credentials have access to testnet/mainnet Terraform state or EC2 instances, a modified workflow file in a PR branch *could* target them (workflow_dispatch runs on the selected branch's version of the workflow). Consider:
- Using an IAM role that can only manage resources tagged `Environment=devnet*`
- Or using OIDC-based auth with conditions on the workflow

### Verdict

**Testnet is safe from the current workflow as-is** — the `devnet-` prefix, config file gating, and Terraform workspace isolation provide layered protection. But the gaps above mean the protection relies on _convention_ rather than _enforcement_. The input sanitization fix is low-effort and should go in with this PR. The IAM scoping is a longer-term hardening item.

dashinfraclaw:
## Follow-up review — latest commit (13:12 UTC)

Nice, the safety checks from my earlier audit are now in. Checking the new additions:

### ✅ Input sanitization added
- Lowercase alphanumeric + hyphens regex — **good**
- Reserved name blocklist (testnet, mainnet, mainnet-*) — **good**
- Double-validation (raw input + resulting devnet-NAME) — belt-and-suspenders, **good**
- Applied to **both** create and destroy workflows — **good**

### ✅ DEVNET_ONLY_GUARD env var
Set on both workflows. Not currently consumed by any downstream script, but serves as a signal/assertion that these workflows are devnet-only. Could be checked in `bin/deploy` or `bin/destroy` as an additional guard in the future.

### One small gap remaining
The blocklist catches `testnet`, `mainnet`, and `mainnet-*` but not `regtest` or `mainnet-support` (which exists in dash-network-configs). Probably fine since those would become `devnet-regtest` / `devnet-mainnet-support` which are harmless, but for completeness:
```bash
if [[ "$NAME" == "testnet" || "$NAME" == "mainnet" || "$NAME" == "regtest" || "$NAME" == mainnet-* ]]; then
```

Otherwise LGTM. The safety story is solid now.

ktechmidas (approved):
LGTM

PR #739, opened 2026-04-02
feat: use per-devnet GPG-encrypted SSH keys in CI workflows

vivekgsharma:
# Summary

This PR introduces **per-devnet SSH access in CI**, replacing the previous shared server key approach with devnet-specific credentials.

# Key Changes

## Unique SSH keypair per devnet
- `Create Devnet` now generates a dedicated SSH keypair for each devnet.
- The private key is encrypted using GPG and stored alongside the devnet config files in:
  - https://github.com/dashpay/dash-network-configs
  - Format: `devnet-<name>.ssh.asc`

## Secure key reuse across workflows
- Subsequent workflows (`Create Resume`, `Manual Deploy`, `Platform Deploy`, `Destroy`) decrypt the stored key in CI using the configured GPG key.
- This ensures consistent, devnet-specific SSH access instead of relying on a shared key.

## Manual access for developers
- Developers can download `devnet-<name>.ssh.asc` from `dash-network-configs`.
- Using their GPG private key, they can decrypt it locally to access and inspect devnet servers.

# Fixes

## Dashmate (fresh-node) issues in Ansible
- Ensures the requested Dashmate version is used for config rendering.
- Always adds the `dashmate` user to the `docker` group.

# Impact

- Improves security by isolating SSH access per devnet  
- Enables reproducible and consistent CI workflows  
- Provides controlled manual access for debugging and inspection


ktechmidas (approved):
LGTM

PR #741, opened 2026-04-15
ansible: codify dashmon status monitoring access

dashinfraclaw:
## Summary
- finish the `status_monitoring` role so it manages the full `dashmon` setup on masternodes
- update the dashboard role to use the dedicated `dashmon` key/user instead of the admin deploy key
- sync the deployed `dashmon-check` script with the current live version, including Tenderdash proposer info

### status_monitoring role
- create the `dashmon` user with a locked password and home directory
- install the dashboard's read-only public key into `/home/dashmon/.ssh/authorized_keys`
- enforce the forced command for that key (`/usr/local/bin/dashmon-check`)
- install `/usr/local/bin/dashmon-check`
- install `/etc/sudoers.d/dashmon`
- add the monitoring public key + sudoers file to the role

### status_dashboard role
- use a dedicated `status_dashboard_ssh_private_key_path` (default `~/.ssh/dashmon-testnet`)
- use `SSH_USER=dashmon`
- stop defaulting to the admin deploy key for status polling

## Verification / provenance
This was reconstructed from the currently running setup rather than guesswork:
- checked the live dashboard host (`68.67.122.250`) and confirmed it is configured with `SSH_KEY_PATH=/home/ubuntu/.ssh/dashmon-testnet` and `SSH_USER=dashmon`
- checked `hp-masternode-3` and confirmed the expected live node-side state:
  - `dashmon` user exists
  - forced-command `authorized_keys` entry matches the dashboard read-only key
  - `/usr/local/bin/dashmon-check` is present
  - `/etc/sudoers.d/dashmon` is present

## Why
`hp-masternode-1` drifted into a bad state for monitoring. Humans could still log in through the normal admin path, but the dashboard uses the dedicated `dashmon` path. The repo already had pieces of this (`status_monitoring`, `dashmon-check`), but not the whole thing, and the dashboard role still defaulted to the wrong key/user.

## Summary by CodeRabbit

* **New Features**
  * Enhanced monitoring system with Tenderdash validator and proposer detection capabilities.
  * Added portable development environment setup with required deployment tools.

* **Chores**
  * Configured dedicated monitoring user with restricted SSH access and sudoers policies.
  * Updated status dashboard configuration to use configurable SSH credentials and user settings.

### DIFF
--- a/ansible/roles/status_dashboard/defaults/main.yml
+++ b/ansible/roles/status_dashboard/defaults/main.yml
@@ -3,5 +3,7 @@
 status_dashboard_image: dashpay/status:latest
 status_dashboard_port: 3010
 status_dashboard_path: "{{ dashd_home }}/status_dashboard"
+status_dashboard_ssh_private_key_path: "{{ lookup('env', 'STATUS_DASHBOARD_SSH_KEY_PATH') | default('~/.ssh/dashmon-testnet', true) }}"
+status_dashboard_ssh_user: dashmon
 status_dashboard_poll_interval: 10000
 status_dashboard_poll_concurrency: 20
--- a/ansible/roles/status_dashboard/tasks/main.yml
+++ b/ansible/roles/status_dashboard/tasks/main.yml
@@ -12,9 +12,9 @@
     dest: "{{ status_dashboard_path }}/inventory"
     mode: "0644"
 
-- name: Copy SSH deploy key for status dashboard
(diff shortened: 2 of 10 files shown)


vivekgsharma:
Superseded by follow-up work on vivek/poll-testnet-status, which includes the status-monitoring changes and the new testnet recovery issue automation.

